PR #27, opened 2015-01-05
Implement caching

djmitche:
Lots of apps need caching for efficiency and to hold down load on other services.

Relengapi should make this easy to get access to.

### DIFF
--- a/docs/deployment/@relengapi/config.rst
+++ b/docs/deployment/@relengapi/config.rst
@@ -164,6 +164,22 @@ If you prefer to supply credentials directly in the RelengAPI configuration, do 
         'secret_access_key': 'secret',
     }
 
+.. _memcached-configuration:
+
+Memcached
+---------
+
+Blueprints may require a memcached instance to cache data.
+There is a common configuration syntax for this.
+
+To use a normal memcached cluster, give the list of `"host:port"` pairs describing the servers::
+
+    SOME_BLUEPRINT_CACHE = ['host-a.foo.com:11211', 'host-a.foo.com:11211']
+
+To use Amazon ElastiCache, give a URL with scheme `elasticache` and the configuration endpoint::
+
+    SOME_BLUEPRINT_CACHE = 'elasticache://mycachecluster2.b47jtf.cfg.use1.cache.amazonaws.com:11211'
+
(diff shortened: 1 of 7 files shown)


djmitche:
OK, it's ready.  @jgmize, want to have a look, since you chimed in?

jgmize:
OK, @djmitche, but be careful what you wish for ;) 

First, let me commend you on your efforts: I can tell you spent a lot of time thinking though your use cases, paying plenty of attention to the edge cases and error conditions. A non-trivial amount of that effort was spent wrapping the non-thread-safe client with the appropriate locks: it looks to me like you have all the necessary locking in place, but as you clearly know, it's easy to miss an edge case. After some bad experiences with multithreading in python, I avoid it altogether whenever practical, and prefer to use multiprocessing or asyncio instead. Of course, since you're already using threading elsewhere in your codebase, that sort of change would be outside the scope of this PR, but you may want to consider it in the future.

Another issue I had was that the number of classes I needed to traverse to fully understand what a given method was doing did feel a bit excessive to me at times, but that is due to my personal preference for a more functional approach-- from an OOP perspective, the code is nicely factored, and I like that you avoided using multiple inheritance.

One final question I have is with regard to the (elasticache_config_server)[https://github.com/mozilla/build-relengapi/pull/27/files#diff-228e3abcece5fe7c0e9d1a78a5c70feeR22] you implemented for testing. It seems to me like that would be better placed in a test suite for the the elasticache-auto-discovery library itself, and your tests would just mock out those calls. You may want to consider forking and submitting a PR against that library.

djmitche:
Great points, thanks.  Yes, threading comes with Flask, so that's life.  I do a lot of work in Twisted Python, too, and that has its own downsides.

The number of classes scared me a little too.  It came about partially because I followed TDD and did a lot of refactoring -- and it was a natural way to decompose the behaviors.  I think that class hierarchies are bad for building APIs (for example, if users are expected to write a subclass to interface with a feature), but in this case the classes are all implementation details of a single module, so I think the result is OK.

And you're absolutely right about mocking out the auto discovery calls, rather than firing up a TCP server.  I'll get that fixed.

Thanks again!

PR #171, opened 2015-01-05
Get better coverage

djmitche:
I was bored..

And yes. the apidoc bit is probably cheating, but I barely understand how docutils works, let alone how to test it.  Besides, the code is exercised by building the docs, and anyway it's not production code.

### DIFF
--- a/coveragerc
+++ b/coveragerc
@@ -21,4 +21,4 @@ exclude_lines =
 
 omit =
     */test_*
-    relengapi/blueprints/docs/temp/build/src/conf.py
+    relengapi/lib/apidoc.py
--- a/relengapi/lib/apidoc.py
+++ b/relengapi/lib/apidoc.py
@@ -26,6 +26,11 @@ from sphinx.util.docfields import GroupedField
 from sphinx.util.docfields import TypedField
 from sphinx.util.nodes import make_refnode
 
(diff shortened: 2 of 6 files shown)


Callek:
Not a fan of the nocover littered here especially in the docutils thing.

Can we do --omit for the docutils piece if need be instead? (at least for now)

I'd *love* the other error conditions to be tested somehow too.

(I'm not blocking on my comments, just not a fan of the approach)

djmitche:
Can you say why you're not a fan?

Callek:
because it feels like this is all stuff we could/should test not just ignore. (I've seen errors that were baffling in that doc code alone over the time I've worked on slaveloan so far as one example)

djmitche:
Nobody that I know understands it well enough to test it.  Docutils is *extremely* obtuse.  And a failure in the doc code (a) will cause validation to fail and (b) won't cause production issues.

The coverage on these just from building the docs is pretty good - there were, I think, 5 missed lines.

I'd use --omit, but there *is* some production code in the file -- the stuff that's not under "no cover" in this patch.

djmitche:
I was thinking `setup` was production code, but it's not - it's only used to set up the sphinx extension during `relengapi build-docs`.  So I switched to omitting.  @Callek ?

Callek:
:+1:

PR #169, opened 2015-01-09
 Add links to source/issue tracker to clobberer footer

djmitche:
To make it easier for people to file issues/contribute :-)

(from @edmorley at https://github.com/mozilla/build-relengapi-clobberer/issues/31)


djmitche:
Good progress today.

screenshot from 2015-01-09 18 11 13

Not the best styling, but not the worst either.  This supports a Repository of Record ("Contribute!") and an issue tracker ("File a Bug!").

djmitche:
OK, ready for review.  I switched things around a bit so that for pages that are part of the 'relengapi' dist, it just shows one section on the "About" menu.

relengapi/lib/introspection.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def _get_relengapi_metadata(dist):
+    req = dist.as_requirement()
+    try:
+        setup_cfg = pkg_resources.resource_stream(req, 'setup.cfg')
+    except Exception:
+        return {}
+    cfg = ConfigParser.RawConfigParser()
+    cfg.readfp(setup_cfg)
```
Callek: since WorkingSet returns entries from sys.path python modules, can we do some magic here to either do a try/except incase setup.cfg is not in the expected format/state OR do some magic to ensure its truely a relengapi dist?

Callek: ...or maybe just doing the _get..._metadata can be keyed off bp.dist in the above code, rather than from an explicit WorkingSet() enumeration?

relengapi/lib/introspection.py
```diff
@@ -0,0 +1,61 @@
...
+        _distributions = {}
+        for dist in pkg_resources.WorkingSet():
+            relengapi_metadata = _get_relengapi_metadata(dist)
+            if relengapi_metadata:
+                dist.relengapi_metadata = relengapi_metadata
+            _distributions[dist.key] = dist
+
+    if not _blueprints:
+        _blueprints = []
+        entry_points = (list(pkg_resources.iter_entry_points('relengapi_blueprints'))
```
Callek: retain the comment from the source-location for this: "# TODO: warn if relengapi_blueprints is used"

Callek:
Love the overall approach, :+1: from me, if you either can either convince me in IRC that my suggested changes are bad, get someone else to agree with you that my suggested changes are bad (and document why) or make said changes. --- Lots of options, and I'll be happy to see this land!

Callek:
p.s. did not review CSS or html template stuff, but if your screenshot above is any indication I'm satisfied.

relengapi/lib/introspection.py
djmitche: The comment's invalid - we don't need to do that.

relengapi/lib/introspection.py
djmitche: In your second comment, are you suggesting only getting metadata from things that have relengapi blueprints in them?

Callek: Yes, only getting metadata for things that have relengapi blueprints contained.

djmitche:
How's that?

Callek:
Looks good

djmitche:
Hm, `setup.cfg` isn't included in sdist tarballs, I think. :(

djmitche:
Ah, it's in the tarballs, but it doesn't get installed.  Hrmph.

djmitche:
see #172.

PR #172, opened 2015-01-14
relengapi_metadata is not installed

djmitche:
When a dist is installed, its `setup.cfg` is no longer available.  So all of the handy `relengapi_metadata` in there is no longer available.  Apparently this is just hard!!

### DIFF
--- a/relengapi/lib/introspection.py
+++ b/relengapi/lib/introspection.py
@@ -2,7 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import ConfigParser
 import pkg_resources
 
 _blueprints = None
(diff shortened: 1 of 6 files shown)


djmitche:
This seems pretty simple, doesn't use any weird setuptools functionality, and doesn't rely on finding data in extraneous files.

relengapi/lib/introspection.py
```diff
@@ -32,25 +31,15 @@ def _fetch():
...
-    cfg = ConfigParser.RawConfigParser()
-    cfg.readfp(setup_cfg)
-    if not cfg.has_section('relengapi'):
-        return {}
-
-    return {o: cfg.get('relengapi', o) for o in cfg.options('relengapi')}
+    # look for relengapi metadata for every dist containing a blueprint
+    blueprint_dists = {bp.dist.key: bp.dist for bp in _blueprints}.values()
+    for dist in blueprint_dists:
+        ep = pkg_resources.get_entry_info(dist, 'relengapi.metadata', dist.key)
```
Callek: Love this whole approach better than the prior one, but I am curious if this ```.get_entry_info``` + ```bp.dist``` song-and-dance has additional value over: something like ```list(pkg_resources.iter_entry_points('relengapi.metadata'))``` ?

Callek:
:+1: (good to merge) -- no matter what the answer to my line note is.

relengapi/lib/introspection.py
djmitche: It mirrors the idea you suggested earlier, to only get relengapi metadata for dists containing relengapi blueprints, and that there's exactly one metadata instance per blueprint.

Issue #163, opened 2014-12-16
Much more flexible token authentication


djmitche:
(Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1132957)
# Temporary Tokens

The use-case is a taskcluster scheduler which needs to enable very limited access for a single task, in the form of a token that has a higher likelihood of being revealed (via verbose logging or because it's passed to untrusted try-level code).

These tokens should
 * be issuable primarily via API
 * have a limited lifetime
 * be revokable (when the job ends)
 * specify arbitrary, trusted additional metadata that's available to the API (e.g., "can upload performance statistics for test suite XXX for revision YYY")

# User Tokens

The existing tokens are permanent and not tied to a particular person -- so they don't go away when that person does.

We need another form of token, which
* anyone who can authenticate to RelengAPI can issue,
* stops working when the issuer's account stops working, and
* appears internally as a login by that user (e.g., "Human:dmitchell@mozilla.com")
* otherwise does not expire

# OAuth

The idea here is to allow other apps at Mozilla to programmatically request a subset of a user's permissions on RelengAPI.  Similar to the way lots of apps in the real world connect to users' GitHub accounts.  In fact, the process should be modelled on GitHub as closely as possible, to eliminate user surprise.

On GitHub, application registration looks like this:

![tokens-dev-apps](https://cloud.githubusercontent.com/assets/28673/5455018/bb949356-8505-11e4-8ba4-bbe205a85b6b.png)
![tokens-register](https://cloud.githubusercontent.com/assets/28673/5455017/bb909314-8505-11e4-8b5e-d061811c9c04.png)

The authorization ask from users looks like this:

![tokens-authorize](https://cloud.githubusercontent.com/assets/28673/5455056/222a2c52-8506-11e4-90f6-4cf6026a060b.png)

Once apps are authorized, they are displayed like this:

![tokens-authorized-apps](https://cloud.githubusercontent.com/assets/28673/5455034/ee482ab0-8505-11e4-91e5-0d9422ea8e73.png)

GitHub's docs: https://developer.github.com/v3/oauth/

djmitche:
## Requests

Requests can be authenticated by:

Input | User Type | Permissions | Duration | Notes
------|-----------|-------------|----------|-------
browser cookie | HumanUser | from authz lookup of user | refreshed on every request |
permanent token (`prm`) | TokenUser | perms granted to token | unlimited | revokable
[...]

PR #175, opened 2015-02-18
Add token types `usr`, `prm`, and `tmp`.

djmitche:
This is to the end of #163, but doesn't include OAuth.

Note that this doesn't include anything to actually disable user tokens when the corresponding user disappears -- that has some other prerequisites.  So even with this patch we won't turn on `base.tokens.usr.issue` for all users.  But it's a good stopping point in work on #163 and it's a lot of code to review already.


djmitche:
@jvehent this specifically doesn't include OAuth, but if you have a bit of time to look it over, that'd be great.  The interesting bits for you are probably `relengapi/blueprints/tokenauth/*.py` and the docs.

jvehent:
Ack. I pressume the correct doc is in `development/@relengapi/tokenauth.rst` ?

djmitche:
Correct, although the other bits in the diff will interest you too.

pep8rc
```diff
@@ -1,6 +1,8 @@
 [pep8]
 show-source = yes
 statistics = yes
+# E731: assigning a lambda (good for tests)
+ignore = E731
```
Callek: When used well (and you do here) lambda can be ok, but pep8 forbade it because its hard to read/comprehend at a glance for many many uses. (I've seen many examples of them in the past that have been git/hg blamed as you fwiw).

Because of that I personally prefer NOT ignoring and using def _foo():  rather than lambda.

I won't block on this opinion though....

relengapi/blueprints/tokenauth/types.py
```diff
@@ -0,0 +1,54 @@
...
+    id = wsme.types.wsattr(int, mandatory=False)
+
+    #: not-before time for limited-duration tokens
+    not_before = wsme.types.wsattr(datetime, mandatory=False)
+
+    #: expiration time for limited-duration tokens
+    expires = wsme.types.wsattr(datetime, mandatory=False)
+
+    #: metadata fro limited-duration tokens (arbitrary JSON object)
+    metadata = wsme.types.wsattr(JsonObject(), mandatory=False)
```
Callek: feels odd to need to do JsonObject() here, perhaps make the class _JsonObjectType, and instantiate the object once, in that file?

relengapi/blueprints/tokenauth/types.py
```diff
@@ -0,0 +1,54 @@
...
+    typ = wsme.types.wsattr(
+        wsme.types.Enum(unicode, 'prm', 'tmp', 'usr'),
+        mandatory=False,
+        default='prm')
+
+    #: token ID for revokable tokens
+    id = wsme.types.wsattr(int, mandatory=False)
+
+    #: not-before time for limited-duration tokens
+    not_before = wsme.types.wsattr(datetime, mandatory=False)
```
Callek: this datetime (and others) should have some indication of format for wsme and timezone specified in

Callek: datetime formats supported in terms of input, and output, should be tested for. And if any input allows DST ambiguity we should test for that as well.

relengapi/blueprints/tokenauth/tokenstr.py
```diff
@@ -0,0 +1,48 @@
...
+    except BadData:
+        logger.warning("Got invalid signature in token %r", token_str)
+        return None
+
+    # convert v1 to ra2
+    if claims.get('v') == 1:
+        return {'iss': 'ra2', 'typ': 'prm', 'jti': 't%d' % claims['id']}
+
+    if claims.get('iss') != TOKENAUTH_ISSUER:
+        return
```
Callek: ValueError() here or something other than just a blank return?

relengapi/blueprints/tokenauth/test_util.py
```diff
@@ -0,0 +1,87 @@
...
+from relengapi import p
+from relengapi.blueprints.tokenauth.tables import Token
+
+
+class FakeSerializer(object):
+
+    @staticmethod
+    def prm(id):
+        return FakeSerializer.dumps(
+            {"iss": "ra2", "jti": "t%d" % id, "typ": "prm"})
```
Callek: reading this code, I keep getting tripped up at the abbreviations, they feel utterly cryptic to me. And i fear they will stifle improvements down the line (years) if we don't try and expand them, or at least make them a bit more descriptive.

Before I ask you to do that, is there some requirement of a 3rd party library, or outside expectations that require these ultra-short key/value pairs?

relengapi/blueprints/tokenauth/test_util.py
```diff
@@ -0,0 +1,87 @@
...
+    def loads(data):
+        if data[:3] != 'FK:':
+            raise BadData('Not a fake token')
+        else:
+            return json.loads(data[3:])
+
+# sample tokens
+
+
+prm_json = {
```
Callek: needs comment(s) about keeping _json synced with the def insert_*

relengapi/blueprints/tokenauth/test_util.py
```diff
@@ -0,0 +1,87 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
Callek: this file shouldn't begin with test_ otherwise nose will try to scan it for tests, I'd name it testing_utils or utils_testing or something instead.   And then a comment "This file is only used for tests" or some such.

relengapi/blueprints/tokenauth/test_tokenauth.py
```diff
@@ -27,219 +37,510 @@ def test_route():
...
 
 
 # tests
 
-@test_context.specialize(user=userperms([]))
+
+@test_context.specialize(user=userperms([p.base.tokens.prm.view]))
+def test_root(client):
+    """The Angular UI is served at the root path, but requires view permission"""
+    eq_(client.get('/tokenauth/').status_code, 200)
```
Callek: This doesn't test what the docstring says it does, it merely tests that the page is viewable when you have that perm, NOT that it requires that perm (as in, not testing the load without the .view perm being present)

relengapi/blueprints/tokenauth/test_tokenauth.py
```diff
@@ -27,219 +37,510 @@ def test_route():
...
 
 @test_context.specialize(user=userperms([]))
-def test_issue_token_forbidden(client):
-    """Token issuance requires base.tokens.issue"""
-    eq_(client.get('/tokenauth/tokens').status_code, 403)
+def test_issue_prm_token_forbidden(client):
+    """Issuing a permanent token requires base.tokens.prm.issue"""
+    request = {
+        'permissions': ['test_tokenauth.zig'], 'description': 'More Zig'}
+    print client.post_json('/tokenauth/tokens', request).data
```
Callek: debugging cruft

relengapi/blueprints/tokenauth/types.py
djmitche: Flask does that translation - http://flask.pocoo.org/docs/0.10/api/#flask.json.JSONEncoder

If that ends up creating naive datetimes, then we should file a separate bug to get that fixed.

relengapi/blueprints/tokenauth/test_util.py
djmitche: They are all base64-encded and put into the token.  See the docs and the description of JWT.

relengapi/blueprints/tokenauth/tokenstr.py
djmitche: That's equivalent to `return None`, but being explicit is good.

relengapi/blueprints/tokenauth/test_tokenauth.py
djmitche: Hah, good catch.  It doesn't require any particular permission anymore (you just see an empty list and can't issue anything if you don't have any of the multitude of useful permissions).  The root widget doesn't appear in that case, so such users won't find themselves on a blank page unless they go looking.

pep8rc
djmitche: So, one of the issues with using `def` everywhere is that Python dislikes it if you define the same function twice (e.g., in two branches of a conditional).  With `lambda`, that's perfectly fine.  So this is not just a syntactic change -- it prohibits certain structures.  That's not a big deal (this isn't highly optimized code), but seems like PEP8 overstepping its boundaries.

And as I mentioned in the comments, this pep8 change isn't part of this pull req (I got the wrong base accidentally).

relengapi/blueprints/tokenauth/test_util.py
Callek: Ahh missed that JWT defined these key's. WFM (but still confusing to me, but since its a defined thing I say we keep it, despite that)

relengapi/blueprints/tokenauth/types.py
Callek: The issues around datetime are vast, I really want docs on what format is expected included here, at the least. E.g. "01/01/2014 01:02:03 PDT" or "Wed, Feb 25, 2015 at 14:53 UTC" etc. Lots of options, extreme ambiguity which will matter.

I would also like tests regarding the format's known accepted.

Simplification is possibly best here though, like unix timestamp? (since its based on epoch)

Callek:
On the high level, I'd like tmp tokens revokable somehow, and at least viewable. The "limited" could be years from skimming the code, and we can certainly have a tmp token get compromised in a way that is useful to us to revoke. (it may not be a common feature though)

relengapi/blueprints/tokenauth/types.py
djmitche: To quote a wise man, "then we should file a separate bug to get that fixed."

djmitche:
How, exactly, do you propose to do that without database access?

relengapi/blueprints/tokenauth/test_loader.py
```diff
@@ -0,0 +1,123 @@
...
+        client.get('/test_tokenauth',
+                   headers=[('Authentication', 'Bearer TOK/2/v1')]).data)
+    eq_(auth['permissions'], [])
+
+
+@test_context.specialize(db_setup=insert_prm)
+def test_prm_loader(app):
+    with app.app_context():
+        eq_(loader.prm_loader({'typ': 'prm', 'jti': 't1'}).permissions,
+            set([p.test_tokenauth.zig]))
```
Callek: missing a docstring for what this test is doing

Callek: same for the next two tests

relengapi/blueprints/tokenauth/tables.py
```diff
@@ -0,0 +1,39 @@
...
+            tok.user = self.user
+        return tok
+
+    @property
+    def permissions(self):
+        token_permissions = [p.get(permissionstr)
+                             for permissionstr in self._permissions.split(',')]
+        # silently ignore any nonexistent permissions; this allows us to remove unused
+        # permissions without causing tokens permitting those permissions to fail
+        # completely
```
Callek: I didn't notice a test for this. (Am I wrong)

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -2,109 +2,109 @@
...
 
 def permitted():
-    return permissions.can(p.base.tokens.view,
-                           p.base.tokens.issue,
-                           p.base.tokens.revoke)
+    return permissions.can(
+        p.base.tokens.prm.view,
+        p.base.tokens.prm.issue,
+        p.base.tokens.prm.revoke,
+        p.base.tokens.tmp.issue)
```
Callek: This permitted doesn't account for "can a user issue/view their own tokens" Which is what should be the root widget available as I understand it.

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -2,109 +2,109 @@
...
+    attrs['typ'] = user.claims['typ']
+    if 'nbf' in cl:
+        attrs['not_before'] = tz.utcfromtimestamp(cl['nbf'])
+    if 'exp' in cl:
+        attrs['expires'] = tz.utcfromtimestamp(cl['exp'])
+    if 'mta' in cl:
+        attrs['metadata'] = cl['mta']
+    if 'prm' in cl:
+        attrs['permissions'] = cl['prm']
+    # TODO: client_id
```
Callek: What would a "client_id" be, and if we push this forward as a TODO can you cite an ISSUE in the comment?

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -116,111 +116,200 @@ def root():
...
+    if p.base.tokens.prm.view.can():
+        cond.append(tbl.typ == 'prm')
+    if p.base.tokens.usr.view.all.can():
+        cond.append(tbl.typ == 'usr')
+    elif email and p.base.tokens.usr.view.my.can():
+        cond.append(sa.and_(tbl.typ == 'usr',
+                            tbl.user == email))
+    if not cond:
+        return []
+    cond = sa.or_(*cond)
```
Callek: style nit: I'm not a fan of re-assigning a var to a whole new type inside a function def, makes skimming code harder to comprehend. E.g. would love a rename of cond above to conditions, or cond_sql here, or whatever ;-)  (NOT blocking)

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -116,111 +116,200 @@ def root():
...
+    if typ:
+        cond = sa.and_(cond, tbl.typ == typ)
+
+    q = tables.Token.query.filter(cond)
+    return [t.to_jsontoken() for t in q.all()]
+
+
+token_issuers = {}
+token_issuer = lambda typ: lambda fn: token_issuers.__setitem__(typ, fn)
```
Callek: While lambda is good in tests, I *don't* really like this here, in part because it makes tracebacks harder to comprehend. Please expand this to actual function(s).

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -116,111 +116,200 @@ def root():
...
+        'nbf': nbf,
+        'exp': exp,
+        'prm': perm_strs,
+        'mta': body.metadata,
+    })
+    return types.JsonToken(typ='tmp', token=token,
+                           not_before=body.not_before,
+                           expires=body.expires,
+                           permissions=perm_strs,
+                           metadata=body.metadata)
```
Callek: Without an easy way to revoke, could we do some sort of assertion on a max expires, maybe 1 week for a max for now?

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -116,111 +116,200 @@ def root():
...
+            raise BadRequest("missing %s" % attr)
+
+    # All types have permissions, so handle those here -- ensure the request is
+    # for a subset of the permissions the user can perform
+    requested_permissions = [p.get(a) for a in body.permissions]
+    if None in requested_permissions:
+        raise BadRequest("bad permissions")
+    if not set(requested_permissions) <= current_user.permissions:
+        raise BadRequest("bad permissions")
+
```
Callek: Can you put a comment here with something like WSME rejects unsupported type's due to an enum (since its not immediately clear from reading this functions)

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -116,111 +116,200 @@ def root():
...
     """Get a token, identified by its ``id``."""
-    token_data = Token.query.filter_by(id=token_id).first()
+    token_data = tables.Token.query.filter_by(id=token_id).first()
     if not token_data:
         raise NotFound
+
+    if not can_access_token('view', token_data.typ, token_data.user):
+        raise NotFound
+
     return token_data.to_jsontoken()
```
Callek: Followup Fodder: Should we expose tmp tokens on this endpoint if passed in?

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -116,111 +116,200 @@ def root():
...
-    token_str = body
-    try:
-        token_info = current_app.tokenauth_serializer.loads(token_str)
-    except BadData:
-        raise NotFound
-    if token_info['v'] != TOKENAUTH_VERSION:
-        raise NotFound
-    token_data = Token.query.filter_by(id=token_info['id']).first()
-    if not token_data:
+    (this avoids embedding a token in a URL, where it might be logged)return .
```
Callek: )return .

relengapi/blueprints/tokenauth/loader.py
```diff
@@ -0,0 +1,115 @@
...
+        self.token_data = token_data
+        if authenticated_email:
+            self.authenticated_email = authenticated_email
+
+    def get_id(self):
+        jti = (':' + self.claims['jti']) if 'jti' in self.claims else ''
+        return 'token:%s%s' % (self.claims['typ'], jti)
+
+    def get_permissions(self):
+        return self._permissions
```
Callek: I'm slightly concerned that a valid issued user token, (lets say me), keeps all its previous permissions (say I issue a usr token that can admin tokens), And then I leave moco, now suddenly I no longer have the permissions for said action, but my token stays valid.

Can we somehow XOR against the actual permissions said user has at time of token-load?

relengapi/blueprints/tokenauth/loader.py
```diff
@@ -0,0 +1,115 @@
...
+            return fn
+        return dec
+
+    def __call__(self, request):
+        # extract the token from the headers, returning None if anything's
+        # wrong
+        header = request.headers.get('Authentication')
+        if not header:
+            return
+        header = header.split()
```
Callek: this could traceback if there is no space there. I suggest just using if ' ' in header: header.split() oor using string splicing.  (Ditto on style nit about reusing var names)

relengapi/blueprints/tokenauth/static/tokens.html
```diff
@@ -18,16 +18,30 @@
...
+                                (permanent)
+                            </span>
+                            <span ng-if="token.typ == 'usr'" class="text-info">
+                                {{token.user}}
+                            </span> - {{token.description}}
+                            <span class="text-muted">({{ token.typ }})</span>
+                        </small>
+                        <br />
+                        <span ng-repeat="perm in token.permissions">
+                            <perm>{{perm}}</perm>
```
Callek: This feels like the wrong html tag, unless there is some magic here I don't know of:
data:text/html,<perm>foo</perm><perm>bar</perm>  makes the string "foobar" appear.

relengapi/blueprints/tokenauth/static/tokens.html
```diff
@@ -45,11 +59,20 @@ <h3 class="panel-title">Issue New Token</h3>
...
-                        <div class="pull-right">
-                            <button ng-click="issueToken()" ng-hide="form.$invalid">Issue</button>
+                        <div class="radio" ng-if="can_issue_prm && can_issue_usr">
+                            <label><input ng-model="newtoken.typ" type="radio" name="typ" value="usr" /> User Token</label>
+                            <span class="text-muted">- <small>tied to your account</small></span>
+                        </div>
+                        <!-- always show this radio, even if it's the only choice -->
+                        <div class="radio" ng-if="can_issue_prm">
+                            <label><input ng-model="newtoken.typ" type="radio" name="typ" value="prm" /> Permanent</label>
+                            <span class="text-muted">- <small>not tied to your account (use for permanent infrastructure)</small></span>
```
Callek: Should temp tokens be issuable via UI?

relengapi/blueprints/tokenauth/types.py
Callek: I'm more saying, this code should cite at least 1 example of a suitable string for the API calls. Rather than just "not-before time"  which leaves users wondering "what format is suitable for `time`"

Callek:
Regarding temp tokens, I do feel that "unknown, in the wild, tokens with unknown expiration" should never exist. But if you implement an expire  time that is finite as suggested I won't even come close to blocking on that. One option would be to have a DB table with just temp tokens, that we occassionally GC away the missing rows.

With this comment though, I've looked over all code at least once, and would call my review done as soon as we address the remaining questions/comments

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -2,109 +2,109 @@
...
+    # for user-associated tokens, if the .all permission is set,
+    # the access is fine; otherwise very that the user matches and
+    # the .my permission is set.
+    elif typ in ('usr',):
+        if not p.get('base.tokens.{}.{}.all'.format(typ, access)).can():
+            email = get_user_email()
+            if not email or not user or user != email:
+                return False
+            if not p.get('base.tokens.{}.{}.my'.format(typ, access)).can():
+                return False
```
Callek: Err looking at coverage now...

This line (return False if user can't view their own tokens) is not covered by tests

relengapi/blueprints/tokenauth/__init__.py
djmitche: It's not changing type.

relengapi/blueprints/tokenauth/__init__.py
djmitche: Ah, is this what you meant by "limited"?  I think that makes good sense -- otherwise someone about to get fired for leaking could just issue themselves a 100-year temporary token.

relengapi/blueprints/tokenauth/loader.py
djmitche: All revokable tokens will be automatically revoked by a periodic job when a user leaves.  That's not implemented yet (it requires some LDAP fanciness) but it's a requirement before this is deployed.  That should cover the scenario you describe.

Also, XOR's definitely not the operation you want there :)

relengapi/blueprints/tokenauth/static/tokens.html
djmitche: This definitely isn't a data URL.  I don't know what you mean.

relengapi/blueprints/tokenauth/static/tokens.html
djmitche: Nope.

relengapi/blueprints/tokenauth/loader.py
djmitche: No, split works fine if its delimiter is missing:
```
>>> 'abc'.split()
['abc']
```
I added a test though.

djmitche: Also, a traceback just means a 500, which isn't the worst thing in the world.

djmitche:
OK, I think that addresses everything.

relengapi/blueprints/tokenauth/static/tokens.html
Callek: I'm not sure if we should have https://github.com/mozilla/build-relengapi/pull/175/files#diff-d532cf5f4dfa1af3ce5e6caeeeab17d2R54 (tmp token issue permission) by itself able to expose the root widget then. Agree?

relengapi/blueprints/tokenauth/static/tokens.html
Callek: Was me trying to type html into raw github comments, failed. But I since learned that ```<perm>``` is just magic that angular helps us with, and is fine.

relengapi/blueprints/tokenauth/loader.py
Callek: well, thanks for setting me straight regarding split working! and thanks for the test.

That said, I'd argue that *always* when the app itself is not broken it shouldn't return ISE 500 errors, and instead should do BadRequest and the like. 500 should be a sign of broken code. (not broken input)

relengapi/blueprints/tokenauth/loader.py
Callek: The "I leave MoCo" or "I lose LDAP" is not the only case I was concerned about, what if I change team and lose a perm, but had a token that included it. My token will outlast my actual permission here, I think we should enforce the token doesn't have more perms than the user its associated to has.

With how large this PR is, and how many comments on it, I'm fine if we spin this to a new issue though.

djmitche: It covers that case, too (the token would be disabled)

Callek: Not really, there will be a gap between the badpenny/cron/whatever disabling the token and the user logging losing the permissions. That gap could be "X" long (however often we run the "everything" revoke), I'm looking for some sort of assertion at "token use" that its valid with those perms, outside of the revoke-script.  (we could have it auto-revoke in such a case, or whatever), but as I said this can be a followup.

djmitche: Yes, that gap is OK.  It's not like the HR team presses "enter" and the person's badge stops working within milliseconds.

Callek:
For history purposes, in my own relengapi testing instance, I just updated to post-this-change. and Had the unfortunate issue of "I was logged in" and thus got errors about the base.tokens.view being invalid, and had to force a logout to avoid.

Then I had the issue that I couldn't view the tokenauth page because the relengapi db was not updated. (`auth_tokens` had the old schema) and since we don't have any alembic/DB-migration stuff the fastest for me was to drop table auth_tokens table and recreate it.  BUT in production right now we have two existing tokens: https://api.pub.build.mozilla.org/tokenauth/  that would restrict us doing a simple drop->createdb

Not sure what the right solution is for these but it should probably be articulated before we go live with this version

PR #179, opened 2015-03-03
Add setting for celery logging

Callek:
I had a desire to use debug logging for celery for relengapi, but I wanted it without the added global debugging (using `celery worker --loglevel debug` adds loads of other output)  (my current solution has been `print` which is sub-par)

Basically I just wanted relengapi.* to log in debug mode.  I got the general idea for this approach from http://www.toforge.com/2011/06/celery-centralized-logging/

The structure of it stemmed mostly from needed to have the signal setup before celery fully instantiates (calls its signals), but after the relengapi app is setup, and in a spot that the signal can actually access the app.

This code would need docs and test(s) if it was to be merged in, but I wanted a :+1: for the general idea first.  @djmitche or @mrrrgn . What do you think?

### DIFF
--- a/relengapi/lib/celery.py
+++ b/relengapi/lib/celery.py
@@ -7,17 +7,14 @@ from __future__ import absolute_import
 import logging
 
 from celery import Celery
-from celery.signals import after_setup_logger
+from celery.signals import celeryd_after_setup
 from flask import current_app
 from werkzeug.local import LocalProxy
 
 _defined_tasks = {}
-_relengapi_log_lvl = None  # Used for celery log level
 
 
 def make_celery(app):
-    global _relengapi_log_lvl
-    _relengapi_log_lvl = app.config.get("RELENGAPI_CELERY_LOG_LEVEL", None)
     broker = app.config.get('CELERY_BROKER_URL', 'memory://')
     celery = Celery(app.import_name, broker=broker)
     celery.conf.update(app.config)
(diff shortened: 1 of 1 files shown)


relengapi/lib/celery.py
```diff
@@ -4,14 +4,20 @@
...
 
+import logging
+
 from celery import Celery
+from celery.signals import after_setup_logger
 from flask import current_app
 from werkzeug.local import LocalProxy
 
 _defined_tasks = {}
+_relengapi_log_lvl = None  # Used for celery log level
```
djmitche: I don't like this because it's independent of any particular app.

Does `current_app` work from the `after_setup_logger` signal?  Alternately, can you read `RELENGAPI_CELERY_LOG_LEVEL` from `celery.conf` during that signal hook?

Callek:
@djmitche is this better/acceptable? (if so I can adjust docs and see about testing, in near future.)

djmitche:
Yes, very nice!

djmitche:
@callek -- ping re testing/docs?

djmitche:
@Callek ping?  I'd like to get 2.0.0 out soon, but this still needs testing and docs..

djmitche:
Also, this would be really handy for me in debugging tooltool :)

Callek:
@djmitche fwiw I don't expect to get to the docs/testing part of this, this week. If you want to pick this up and run with it by all means. Otherwise I expect to round it out early in April

PR #183, opened 2015-03-05
Add special case for connecting to S3

djmitche:
The typical boto.connect_s3 method doesn't take a 'region' argument like
most of the others, so we use boto.s3.connect_to_region instead.

### DIFF
--- a/relengapi/lib/aws.py
+++ b/relengapi/lib/aws.py
@@ -35,6 +35,16 @@ class AWS(object):
         if key in self._connections:
             return self._connections[key]
 
+        # handle special cases
+        try:
+            fn = getattr(self, 'connect_to_' + service_name)
+        except AttributeError:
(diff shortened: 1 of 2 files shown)


djmitche:
@mrrrgn how's this look to you?

djmitche:
I'd like to release relengapi-1.1.1 with just this change, since without it relengapi-tooltool's tests fail.  But there's stuff on master that I don't want to release just yet as it requires schema and configuration changes.

Callek:
:+1: Note: I support the "release ... with this" if we don't munge history here (e.g. don't revert stuff on master, and make sure this merges in with relnotes and such)

djmitche:
I was planning to cherry-pick onto the relengapi-1.1.0 branch and make the
release, then merge that into master.

Issue #132, opened 2014-10-10
Badpenny: locking support


djmitche:
As Badpenny starts to replace crontasks, we'll want a way to make sure jobs don't overlap one another.  This can just be an extra argument (or better, decorator) for tasks.

The tricky bit will be making sure that jobs don't get "wedged" when, for example, a celery host fails or a DB write fails.  Ideally we could tie this down to TCP connections, perhaps with DB locks?  Then if a job fails for whatever reason, eventually that TCP connection comes down and some periodic cleanup task (or a subsequent acquire operation on the lock) can mark it as crashed.

PR #186, opened 2015-03-11
Disable user tokens

djmitche:
@callek, @jvehent 

This, along with #175, should finish off the user-token work described in #163.

### DIFF
--- a/relengapi/blueprints/tokenauth/__init__.py
+++ b/relengapi/blueprints/tokenauth/__init__.py
@@ -78,6 +78,9 @@ def user_to_jsontoken(user):
     if 'prm' in cl:
         attrs['permissions'] = cl['prm']
 
+    # we never load disabled users, so this one isn't disabled
+    attrs['disabled'] = False
+
     if user.token_data:
         td = user.token_data
         attrs['id'] = td.id
@@ -172,7 +175,8 @@ def issue_prm(body, requested_permissions):
     token_row = tables.Token(
         typ='prm',
         description=body.description,
-        permissions=requested_permissions)
+        permissions=requested_permissions,
+        disabled=False)
     session.add(token_row)
     session.commit()
 
@@ -208,7 +212,8 @@ def issue_tmp(body, requested_permissions):
                            not_before=tz.utcfromtimestamp(nbf),
                            expires=body.expires,
                            permissions=perm_strs,
-                           metadata=body.metadata)
+                           metadata=body.metadata,
+                           disabled=False)
 
 
 @token_issuer('usr')
@@ -223,7 +228,8 @@ def issue_usr(body, requested_permissions):
         typ='usr',
         user=email,
         description=body.description,
-        permissions=requested_permissions)
+        permissions=requested_permissions,
+        disabled=False)
     session.add(token_row)
     session.commit()
 
--- a/relengapi/blueprints/tokenauth/__init__.py
+++ b/relengapi/blueprints/tokenauth/__init__.py
@@ -22,7 +22,6 @@ from relengapi.blueprints.tokenauth import tokenstr
 from relengapi.blueprints.tokenauth import types
 from relengapi.lib import angular
 from relengapi.lib import api
-from relengapi.lib import auth
 from relengapi.util import tz
 from werkzeug.exceptions import BadRequest
 from werkzeug.exceptions import Forbidden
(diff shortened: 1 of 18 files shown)


djmitche:
Any comments on this?

Callek:
(I didn't realize this was pending review) that said, I'm curious the use case for getting a users permissions outside of a request context....

djmitche:
The use is in this patch, in fact -- disable user tokens when the user no longer has the permissions embodied in the token.

djmitche:
I'd very much like to deploy this soon, since we'll need user tokens for tooltool uploads.  Any chance of an r?

relengapi/blueprints/tokenauth/usermonitor.py
```diff
@@ -0,0 +1,41 @@
...
+                print "C"
+                disable = False
+
+        if disable and not token.disabled:
+            job_status.log_message("Disabling token %d for user %s" % (token.id, token.user))
+            token.disabled = True
+        elif not disable and token.disabled:
+            job_status.log_message("Re-enabling token %d for user %s" % (token.id, token.user))
+            token.disabled = False
+    session.commit()
```
Callek: nit: this task is scheduled for an hour now, but I am slightly concerned that there will be a time where this takes longer than an hour. And then we have two copies doing duplicate work (the session as it stands means we'll have the same results each cycle).

My ideal would be something like, but it also may be overkill for this use case, I leave that up to you.
```python
user_ids = [tok.user for tok in tables.Token.query.filter(tables.Token.typ == 'usr')]
for id in user_ids:
    session = current_app.db.session('relengapi')
    token = session.query(tables.Token).get(id)
    if not token:
      continue  # if the token was deleted while this was executing, do nothing.
   # ...
   session.commit()
```

relengapi/blueprints/tokenauth/usermonitor.py
```diff
@@ -0,0 +1,41 @@
...
+        print token_perms, user_perms
+        if user_perms is None:
+            print "A"
+            disable = True
+        else:
+            if token_perms - user_perms:
+                print "B"
+                disable = True
+            else:
+                print "C"
```
Callek: A, B, C are debugging cruft I suspect, can you make this either a proper log entry, or remove them?

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -208,7 +212,8 @@ def issue_tmp(body, requested_permissions):
                            not_before=tz.utcfromtimestamp(nbf),
                            expires=body.expires,
                            permissions=perm_strs,
-                           metadata=body.metadata)
+                           metadata=body.metadata,
+                           disabled=False)
```
Callek: (Still reading -- but) Does a tmp token "disabled" mean anything? considering we don't store any metadata that would allow us to "disable" it?

relengapi/blueprints/tokenauth/test_usermonitor.py
```diff
@@ -0,0 +1,125 @@
...
+        def side_effect(user):
+            if user in permissions:
+                return set(permissions[user])
+        gup.side_effect = side_effect
+        yield
+
+
+def assert_disabled(app, js):
+    eq_(tables.Token.query.first().disabled, True)
+    js.log_message.assert_called_with("Disabling token 2 for user me@me.com")
```
Callek: I'm not a big fan of ".first()" with these queries, they don't evoke "this is always what we need/want" to me. (as in, if we in the future change the Token table for these tests for any reasons, stuff like this could break)

Callek:
Only 1 line is missing from tokenauth/__init__ imho might as well test it (at least partly);

https://coveralls.io/builds/2172898/source?filename=relengapi%2Fblueprints%2Ftokenauth%2F__init__.py#L49

That said, if you address my nits :+1: feel free to ping me here if you want a second pass at any new csets. (or need me for any followup to my comments)

relengapi/blueprints/tokenauth/usermonitor.py
djmitche: As I understand it, committing a session with an iterator open will cause that iteration to fail.

I'm not *too* worried about overlaps -- we might get exceptions if both runs try to make changes, but the operations are idempotent so that won't hurt anything.  The real fix is to use badpenny to run no more than one instance at a time (#132).

relengapi/blueprints/tokenauth/__init__.py
djmitche: No, but in the interest of returning a full JSON object it's included.  The token isn't disabled -- that it *can't* be disabled is neither here nor there..

relengapi/blueprints/tokenauth/test_usermonitor.py
djmitche: It's OK for tests to be a little fragile.  That said, `insert_usr` always inserts with id 2, so I'll add `.filter_by(id=2)` here, so at least we're not relying on there being only one row.

PR #187, opened 2015-03-12
prohibit open redirects

djmitche:
I'm open to other ideas for module name beyond 'safety'.  But I figure we'll end up with a half-dozen random utility functions, similar to [Werkzeug](http://werkzeug.pocoo.org/docs/0.10/utils/)

### DIFF
--- a/docs/development/@relengapi/safety.rst
+++ b/docs/development/@relengapi/safety.rst
@@ -1,0 +1,26 @@
+Safety Utilities
+================
+
+.. py:module:: relengapi.lib.safety
+
+RelengAPI is a security-sensitive application, so it's important to avoid common mistakes.
+While Werkzeug provides some utilities to help, this module supplements those with a few RelengAPI-specific tools.
+
+Redirect URLs
+-------------
+
+RelengAPI must not expose any "open redirects", URLs which could result in an HTTP redirect with a URL of a attacker's choosing.
+To avoid this, it is safest to not redirect to a URL provided by the user agent.
+However, sometimes such redirects are convenient, especially after authentication.
+In these cases, the next best approach is to refuse any redirects outside of the RelengAPI service itself.
+
+.. py:function:: safe_redirect_path(url)
+
+    :param url: the potentially dangerous URL
+    :returns: a URL guaranteed to be within the RelengAPI service
+
+    This method will return a relative path unchanged.
+    For URLs with a network location or schema, it will return a path to the root of the RelengAPI service.
+    Use it like this::
+
+        return redirect(safe_redirect_path(url_from_user_agent))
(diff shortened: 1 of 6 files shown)


Callek:
I could be wrong but if we pass url_for to this it looks like it would give us merely "root" not the url we want. (thats a 30 second glance)

djmitche:
Please give it more than a 30 second glance

relengapi/lib/safety.py
```diff
@@ -0,0 +1,16 @@
...
+
+import urlparse
+
+from flask import url_for
+
+
+def safe_redirect_path(url):
+    parts = urlparse.urlparse(url)
+    if parts.scheme or parts.netloc:
+        return url_for('root')
```
Callek: indeed the parts.scheme will fail for any allowed url_for entry. (such as if we ever want to allow the login to default to something other than root, e.g. ```next_url = request.args.get('next') or url_for('root')```

I'd argue we should do something like the following, instead (note completely untested):
```from flask import request, current_app

if request: # Support contexts without a request object, e.g. celery
    if url.startswith(request.url_root):
        return url
    if current_app.config.get("SERVER_ROOT", None) and url.startswith(current_app.config.get(SERVER_ROOT)):
        return url
parts = urlparse.urlparse(url)
    if parts.scheme or parts.netloc:
         return url_for('root')
return url
```

But the idea here is that a url with a scheme+base that is equal to ours should be safe, right?

This of course assumes we trust a url redirect for any url under the relengapi service, is that the case?

djmitche: How will parts.scheme fail?

djmitche: @Callek ping?  Note that `url_for` generates, by default, a relative URL.  Flask converts that to an absolute URL when it's returned in a 30x redirect.  So as far as I can tell this works.

Callek:
I'd like to test some variants before I :+1: would be a few days though. If this actively blocks you feel free to re-ping me, or convince someone else to give it a review.

djmitche:
Sure, there's no great hurry here.

docs/development/@relengapi/safety.rst
```diff
@@ -0,0 +1,26 @@
...
+.. py:function:: safe_redirect_path(url)
+
+    :param url: the potentially dangerous URL
+    :returns: a URL guaranteed to be within the RelengAPI service
+
+    This method will return a relative path unchanged.
+    For URLs with a network location or schema, it will return a path to the root of the RelengAPI service.
+    Use it like this::
+
+        return redirect(safe_redirect_path(url_from_user_agent))
```
Callek: /home/jwood/relengapi/relengapi-docs/development/@relengapi/safety.rst:: WARNING: document isn't included in any toctree

relengapi/lib/safety.py
Callek: hrm, that apparantly seems true.

This safe code assumes (afaict) the ```http(s)://server:port/foo``` means that foo is under the app, and we're not hosted in a subdir.  I think thats fine but worth noting here.

Callek:
:+1: if you fix the doc issue I just found

PR #192, opened 2015-03-20
HTTP header with authentication data is called "Authorization"

djmitche:
..not "Authentication".

### DIFF
--- a/relengapi/blueprints/tokenauth/test_loader.py
+++ b/relengapi/blueprints/tokenauth/test_loader.py
@@ -16,28 +16,28 @@ from relengapi.blueprints.tokenauth.util import insert_usr
 
 @test_context
 def test_loader_no_header(app, client):
-    """With no Authentication header, no permissions are allowed"""
+    """With no Authorization header, no permissions are allowed"""
     auth = json.loads(client.get('/test_tokenauth').data)
     eq_(auth['permissions'], [])
 
 
 @test_context
 def test_loader_not_bearer(app, client):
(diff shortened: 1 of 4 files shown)


djmitche:
@petemoore - do we need to worry about this with the existing mapper stuff?  I can include a backward-compatibility shim to still accept "Authorization" if so, but if those tokens aren't actively being used I'd rather just change the mapper client at the same time.

djmitche:
I added the backward compatibility, just to be safe.  So this isn't quite as much a "breaking change" as originally advertised.

Callek:
Needs a test to exercise the backwards compat functionality. (Thats the reason coverage fell)

relengapi/blueprints/tokenauth/test_loader.py
```diff
@@ -16,28 +16,39 @@
...
     tok = FakeSerializer.prm(1)
     auth = json.loads(
         client.get('/test_tokenauth',
-                   headers=[('Authentication', 'Bearer ' + tok)]).data)
+                   headers=[('Authorization', 'Bearer ' + tok)]).data)
+    eq_(auth['permissions'], ['test_tokenauth.zig'], auth)
+
+
+@test_context.specialize(db_setup=insert_prm)
+def test_loader_good_header_Authentication(app, client):
```
djmitche: Whoops, I wrote the test but used the new header :)

djmitche:
fixed

djmitche:
That was the only comment, so I'm assuming this is mergeable.

Callek:
:+1:

PR #189, opened 2015-03-23
badpenny `cleanup_old_jobs` fails with integrity error

djmitche:
```
IntegrityError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely) (IntegrityError) (1451, 'Cannot delete or update a parent row: a foreign key constraint fails (`relengapi`.`badpenny_job_logs`, CONSTRAINT `badpenny_job_logs_ibfk_1` FOREIGN KEY (`id`) REFERENCES `badpenny_jobs` (`id`))') 'DELETE FROM badpenny_jobs WHERE badpenny_jobs.id = %s' ((1235L,), (1238L,), (1240L,), (1243L,), (1245L,), (1248L,), (1250L,), (1253L,)  ... displaying 10 of 14 total bound parameter sets ...  (1265L,), (1268L,))
[2015-03-18 11:19:27,045: ERROR/Worker-2] Job 1276 failed
```

### DIFF
--- a/relengapi/blueprints/badpenny/cleanup.py
+++ b/relengapi/blueprints/badpenny/cleanup.py
@@ -27,6 +27,8 @@ def cleanup_old_jobs(job_status):
         jobs = reversed(task.jobs[1:])
         for job in jobs:
             if job.created_at < old:
+                for log in job.logs:
+                    session.delete(log)
                 session.delete(job)
                 deleted += 1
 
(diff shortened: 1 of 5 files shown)


djmitche:
So first, why didn't this get caught in testing?  Because SQLite doesn't enforce foreign key consistency by default.  So I enabled that and fixed the bugs it uncovered.

Also because we weren't testing for it.  So I added tests.  Then I fixed the bug!

I'm not sure why the relengapi-docs thing bit me while I was testing this, but I think the fix is sensible anyway.

Callek:
:+1:

PR #194, opened 2015-03-24
Use the 'classic' sphinx theme

djmitche:
Since we present docs inline using Jinja2 templates, this doesn't
actually have much effect, but it silences a warning from sphinx.


Callek:
:+1:

PR #196, opened 2015-03-25
fix JS usage on chrome

djmitche:
It appears that `[].slice.apply` is the normal way to do this anyway.


djmitche:
And `Array.find` [isn't standardized yet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find).  I only needed `Array.some` anyway (and it [is standardized](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some))

Callek:
I'm not great at JS, but :+1: stamp anyway.

PR #199, opened 2015-03-26
include an example of a token description

djmitche:
@amyrrich r?


Callek:
Why not placeholder="..." in the input field instead?

djmitche:
Oh, good idea, thanks!

djmitche:
..updated

Callek:
:+1:

PR #201, opened 2015-03-27
Add logging of token creation and revocation

djmitche:
This adds the logging mentioned in #163.

### DIFF
--- a/relengapi/blueprints/tokenauth/__init__.py
+++ b/relengapi/blueprints/tokenauth/__init__.py
@@ -276,6 +276,8 @@ def issue_token(body):
     if not set(requested_permissions) <= current_user.permissions:
         raise BadRequest("bad permissions")
 
+    logger.info("Issuing token to {} with permissions {}".format(
+        current_user, requested_permissions))
     # Dispatch the rest to the per-type function.  Note that WSME has already
     # ensured `typ` is one of the recognized types.
     return token_issuers[typ](body, requested_permissions)
@@ -336,6 +338,8 @@ def revoke_token(token_id):
 
     if not can_access_token('revoke', token_data.typ, token_data.user):
         raise Forbidden
+
+    logger.info("Revoking {} token #{}".format(token_data.typ, token_data.id))
 
     tables.Token.query.filter_by(id=token_id).delete()
     session.commit()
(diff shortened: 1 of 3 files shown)


relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -276,6 +276,8 @@ def issue_token(body):
     if not set(requested_permissions) <= current_user.permissions:
         raise BadRequest("bad permissions")
 
+    logger.info("Issuing token to {} with permissions {}".format(
+        current_user, requested_permissions))
```
Callek: current_user.authenticated_email ?

relengapi/blueprints/tokenauth/__init__.py
```diff
@@ -337,6 +339,8 @@ def revoke_token(token_id):
     if not can_access_token('revoke', token_data.typ, token_data.user):
         raise Forbidden
 
+    logger.info("Revoking {} token #{}".format(token_data.typ, token_data.id))
+
```
Callek: worth logging which perms are revoked, rather than a token id?  (if thats not hard)

relengapi/blueprints/tokenauth/usermonitor.py
```diff
@@ -25,10 +29,16 @@ def monitor_users(job_status):
                 disable = False
 
         if disable and not token.disabled:
-            job_status.log_message("Disabling token %d for user %s" % (token.id, token.user))
+            logmsg = "Disabling token %d for user %s" % (
+                token.id, token.user)
```
Callek: same idea about perms and not id.

Callek:
:+1:

relengapi/blueprints/tokenauth/__init__.py
djmitche: No, that will raise AttributeError if there's no email associated with the authentication (e.g., a token)

relengapi/blueprints/tokenauth/__init__.py
djmitche: The whole token is revoked, so I'm not sure how useful that would be..

djmitche:
Updated -- better?

Callek:
:+1: with the fix mentioned

PR #184, opened 2015-03-27
Make sure "Flask App" docs include all app attributes

djmitche:
https://api.pub.build.mozilla.org/docs/development/@relengapi/app/ probably doesn't list everything it should.

### DIFF
--- a/docs/development/@relengapi/app.rst
+++ b/docs/development/@relengapi/app.rst
@@ -5,20 +5,47 @@ The RelengAPI Flask App is mostly a normal Flask App, but has a few additional a
 
 .. py:class:: flask.Flask
 
+    .. py:attribute:: authz
+
+        Access to authorization functionality outside of a request context.
+        See :doc:`auth`.
+
+    .. py:attribute:: aws
+
+        Amazon Web Services functionality.
+        See :doc:`aws`.
+
+    .. py:attribute:: celery
+
+        A [Celery](http://www.celeryproject.org/) app -- for internal use.
+        See :doc:`tasks`.
+
+    .. py:attribute:: db
+
+        Database access.
+        See :doc:`databases`.
+
+    .. py:attribute:: layout
+
+        Control of the global Jinja2 layout.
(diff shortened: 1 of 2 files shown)


docs/development/@relengapi/app.rst
```diff
@@ -5,20 +5,47 @@ The RelengAPI Flask App is mostly a normal Flask App, but has a few additional a
...
+
+        This attribute gives a list of all RelengAPI blueprints.
+        It is different from ``current_app.blueprints`` in that it does not include blueprints from Flask extensions.
+        Each RelengAPI blueprint has a ``dist`` attribute giving the SetupTools distribution from which the blueprint came.
 
-    .. py:attribute:: authz
 
-        See :doc:`auth`
+Blueprints can add their own attributes to the application as necessary.
+These should generally begin with the name of the blueprint to avoid name collisions.
```
Callek: would be great if we could come up with a way to have apps hook docs into here (with something like ```[from relengapi-mapper]``` annotation) I don't think thats easy at present so not blocking

docs/development/@relengapi/app.rst
```diff
@@ -5,20 +5,47 @@ The RelengAPI Flask App is mostly a normal Flask App, but has a few additional a
...
+
+    .. py:attribute:: memcached
+
+        Access to configured memcached caches.
+        See :doc:`memcached`
+
+    .. py:attribute:: relengapi_blueprionts
+
+        Access to configured memcached caches.
+        See :doc:`memcached`
```
Callek: duplicated attribute "relengapi_blueprints" (with typo) but listing docs for memcached. I suspect this is either "needs to be deleted" or you meant to list something else here.

Callek:
:+1: once comments addressed

docs/development/@relengapi/app.rst
djmitche: I'm not sure that's possible at all with sphinx, tbh..

PR #205, opened 2015-04-01
Celery logging

djmitche:
#179, but with tests and docs.  r? @Callek?


Callek:
Not familiar with that method of log testing, but I like it. Thanks for the docs/tests :+1:

djmitche:
Well, I made it up  -- I had to pipe the logs back from the subprocess
somehow!

Thanks

PR #173, opened 2015-01-28
Patch for issue #48 Enable logging of SQL statements 

rahulrrixe:
@djmitche  another way to do this is. Let me know which would be preferred.

```Python
import logging

logging.basicConfig()
logging.getLogger('sqlalchemy.engine').setLevel(logging.INFO)
```

Issue #48, opened 2014-05-01
Enable logging of SQL statements


djmitche:
It's sometimes useful to see all of the SQL queries that things issue in the logs.

SQLAlchemy supports this with an `echo=True` argument to engine construction.  We don't want to turn this on all the time, though!

djmitche:
@rahulrrixe is working on this in #173.

PR #212, opened 2015-04-03
Enable logging of SQL statements

djmitche:
This solves #48, using @rahulrrixe's commit from #173.

Rahul, what do you think?


djmitche:
@Callek in Rahul's absence, r?

rahulrrixe:
@djmitche Is more changes required in this?

djmitche:
I don't think so -- I was asking you if the result looked good.  Do you see anything else that isn't complete?

Callek:
Looks ok to me, I didn't test though. :+1:

rahulrrixe:
@djmitche It is woking fine on my local system. :+1:

PR #213, opened 2015-04-07
Bug 1151636: ignore results for badpenny tasks

djmitche:
This avoids creating a lot of result queues that serve no purpose.


Callek:
:+1:

PR #215, opened 2015-04-07
Badpenny tasks are always shown as finishing "instantly"

djmitche:
This row from badpenny_jobs:
```
| 319 |       4 | 2015-04-07 19:15:05 | 2015-04-07 19:15:05 | 2015-04-07 19:23:49 |          0 | NULL   | NULL |
```
displays as "Job started 4/7/2015 3:15:05 PM, finished instantly (details)".  That ain't right.


djmitche:
We're still only getting one-second resolution (#216) but at least this is accurate to that scale.  r?

Callek:
stamp :+1:

PR #217, opened 2015-04-08
Add 'relengapi repl' for debugging use

djmitche:
I found this useful, and it doesn't hurt.

### DIFF
--- a/relengapi/blueprints/base/__init__.py
+++ b/relengapi/blueprints/base/__init__.py
@@ -85,3 +85,25 @@ class SQSListenSubcommand(subcommands.Subcommand):
 
     def run(self, parser, args):
         current_app.aws._spawn_sqs_listeners()
+
+
+class ReplSubcommand(subcommands.Subcommand):
+
+    def make_parser(self, subparsers):
+        parser = subparsers.add_parser(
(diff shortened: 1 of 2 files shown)


relengapi/blueprints/base/__init__.py
```diff
@@ -85,3 +85,23 @@ def make_parser(self, subparsers):
...
+        return parser
+
+    def run(self, parser, args):
+        import code
+        # try to get readline, but don't fail if it's not available
+        try:
+            import readline
+            assert readline
+        except ImportError:
+            readline = None
```
Callek: ...but where is readline module/var used?

djmitche: The interpreter uses it if it's present

djmitche:
comment updated

relengapi/blueprints/base/__init__.py
Callek: Ahh my missing piece was the fact that readline does this with input (and raw_input) https://docs.python.org/2/library/functions.html?highlight=input#input

Callek:
:+1: I've done this with flask's debug mode enabled and simply raising an exception during a request. But having this for cases where I don't need a request context will be useful

PR #221, opened 2015-04-17
allow RELENGAPI_SETTINGS to be relative to the current directory

djmitche:
Without this, I have to manually set RELENGAPI_SETTINGS when I switch between blueprints (tooltool vs. clobberer vs. relengapi).  With this, I can just set it to `settings.py` and it will use that file in the cwd.

r?


Callek:
:+1:

PR #222, opened 2015-04-20
Implement 'constant' auth

djmitche:
This is handy when you're not on the 'net and thus browserid is not
practical.


djmitche:
@Callek r?

Callek:
NIT I'd recommend a WARNING that constant auth is unsafe, since anyone who can access is mimicking whatever user you specify.  Wording up to you.

:+1: though.

djmitche:
It does say it's for development..

Callek:
...meant as a log warning, not just docs warning.

PR #190, opened 2015-04-20
API method to force a run of a badpenny task

djmitche:
For testing and probably a lot of other reasons, it'd be nice to be able to manually trigger a run of a badpenny task.

This should be a single new API method, protected by a new permission.


djmitche:
OK, r?  I fixed #211 and #209 while I was at it.

Callek:
:+1: if you acknolwedge that I didn't review the angular/html changes.

PR #69, opened 2015-04-21
Fold validate.sh output in Travis

djmitche:
Travis sets a few environment variables, so we can tell when we're in travis:
```
TRAVIS_JOB_ID=25905243
```
and in that case, issuing lines like
```
travis_fold:start:something
...
travis_fold:end:something
```
will fold the contained contents down to a single line.


djmitche:
r?

If this works, I'll do the same in skeleton and merge down to the blueprints.

Callek:
:+1: though commit message is wrong now that you've pulled out the "buffer otherwise" part (as I requested -- thanks)

PR #188, opened 2015-04-21
During tests, log all paths used with the test client

djmitche:
So `client.get('/foo/bar')` should log something like
```
test_client: INFO: GET /foo/bar
```

This will require monkey-patching werkzeug a little bit; see https://github.com/mozilla/build-relengapi/blob/master/relengapi/lib/testing/context.py#L85


djmitche:
r?

Callek:
:+1:

PR #161, opened 2015-04-22
Improve the <perm> directive to describe permissions in a tooltip

djmitche:
If you hover over a permission anywhere in the Releng API UI, it should give you the description for that permission.

This will require adding an endpoint that describes permissions (probably in the `auth` blueprint), then calling that appropriately from Angular code.


djmitche:
r?

Callek:
:+1:

PR #33, opened 2015-04-22
Use python-versioneer to handle version numbers

djmitche:
@warner has a cool library that makes it easier to bump versions of Python packages.  RelengAPI should use it, and recommend that other blueprints use it too.


djmitche:
I can't get validate.sh to run successfully.  Two problems.  First, the version changes every time you commit, which causes python to be sad:
```
-- building docs --
Traceback (most recent call last):
  File "/home/dustin/code/relengapi/t/relengapi/sandbox/bin/relengapi", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/home/dustin/code/relengapi/t/relengapi/sandbox/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3038, in <module>
    working_set = WorkingSet._build_master()
  File "/home/dustin/code/relengapi/t/relengapi/sandbox/lib/python2.7/site-packages/pkg_resources/__init__.py", line 617, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/home/dustin/code/relengapi/t/relengapi/sandbox/lib/python2.7/site-packages/pkg_resources/__init__.py", line 630, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/home/dustin/code/relengapi/t/relengapi/sandbox/lib/python2.7/site-packages/pkg_resources/__init__.py", line 808, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: relengapi===2.0.3-19-gd995389
** build-docs failed **
```
and secondly, `python setup.py sdist` adds a `MANIFEST` file and then does not include any of the files listed in that manifest in the sdist file.

I'm not sure why either of these are the case, but I've now put more time into this than it might save.

PR #203, opened 2015-04-25
On token generation page, "issue" button should be disabled, not hidden

djmitche:
Mostly a new user issue - it's not obvious that to issue a token, the description must be non-empty.
Having the "issue" button visible, but disabled, would suggest that something more needs to be done.


djmitche:
@hwine r?

hwine:
+1 lgtm

hwine:
er :+1: I meant

PR #210, opened 2015-04-25
Format user tokens better

djmitche:
When authenticating with a token, str(current_user) contains only the token ID, which doesn't indicate *whose* token it is.  That info should be readily avaliable, and would make logging a lot more helpful!


djmitche:
r?

Callek:
:+1:

PR #31, opened 2015-05-01
Handle extra MySQL engine options

djmitche:
When setting up a MySQL engine or connection, set some extra options:
* storage_engine (defaulting to InnoDB)
* always use unicode
* set timezone = UTC

Buildbot is a good example for the storage engine, at least (although it sets MyISAM).

### DIFF
--- a/relengapi/lib/db.py
+++ b/relengapi/lib/db.py
@@ -117,6 +117,20 @@ class Alchemies(object):
 
         return engine
 
+    def create_mysql_engine(self, url):
+        url.query['use_unicode'] = "True"
+        engine = sa.create_engine(url)
+
+        @event.listens_for(engine, "connect")
+        def foreign_keys_on(dbapi_con, con_record):
+            # Set the default storage engine in case we create tables
+            dbapi_con.cursor().execute("SET default_storage_engine=InnoDB")
+            # and set the timezone for this connection to avoid any undesired
+            # translation
+            dbapi_con.cursor().execute("SET time_zone=UTC")
+
+        return engine
+
     @synchronized(threading.Lock())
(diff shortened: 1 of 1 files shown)


djmitche:
I use mysql locally for testing, and this works OK.  Still, we'll want to be careful deploying this in production to make sure that timestamps aren't read back incorrectly or anything funny like that.

djmitche:
@Callek since you've been concerned with UTC quite a bit, mind having a look?

relengapi/lib/db.py
```diff
@@ -91,19 +92,45 @@ def engine(self, dbname):
...
+
+        @event.listens_for(engine, "connect")
+        def foreign_keys_on(dbapi_con, con_record):
+            # Set the default storage engine in case we create tables
+            dbapi_con.cursor().execute("SET default_storage_engine=InnoDB")
+            # and set the timezone for this connection to avoid any undesired
+            # translation
+            dbapi_con.cursor().execute("SET time_zone=UTC")
+
+        return engine
```
Callek: So, I'm slightly confused here. Help me to understand.

Does the old function continue to operate after relengapi reloads?
Do we want `PRAGMA foreign_keys = ON` enabled for MySQL?

I think both SET's are a good choice for MySQL at the least

djmitche: By "old function" I think you mean the listener that sets "PRAGMA foreign_keys = ON".  Yes, it does continue to operate after relengapi reloads.

No, we don't want `PRAGMA foreign_keys = ON` for MySQL, because it's not a valid MySQL statement.

Callek:
:+1:

PR #228, opened 2015-05-07
Add support for Mozilla-style POSIX groups

djmitche:
This will enable us to match scm_level_* groups, specifically.

@jdow does this look awful?  @callek, designed review-guy-for-relengapi, what do you think?


jdow:
That looks good to me.

Callek:
this is pretty much a rubber-stamp, but with jdow looking at it I have no concerns

:+1:

Issue #229, opened 2015-05-08
MySQL doesn't like "UTC"


djmitche:
In #31 we added the ability to set the per-connection timezone.  Worked on my local system, but not on MySQL:
```
OperationalError: (_mysql_exceptions.OperationalError) (1298, "Unknown or incorrect time zone: 'UTC'")
```

djmitche:
```
11:21:19 <sheeri> So I would use utc_timestamp() to get the values, and store them in datetimes.
11:21:38 <sheeri> the value of the server's timezone does not affect the UTC_TIMESTAMP() function
11:21:51 <sheeri> but it does affect CURRENT_TIMESTAMP() and NOW()
```

so I think, for now, we should back out the "set time_zone" change, and separately make efforts to avoid timezone-sensitive functions and columns.

djmitche:
Also, per http://technocation.org/files/doc/2009_12_timezones.pdf only TIMESTAMP columns are sensitive to the server-side timezone, so if we don't use that column type then we don't have to worry about this.

djmitche:
Also, in trying to insert new data from treestatus:
`[Fri Aug 28 12:51:08 2015] [error] OperationalError: (_mysql_exceptions.OperationalError) (1292, "Incorrect datetime value: '2015-08-28 19:51:08.081088+00:00' for column 'when' at row 1") [SQL: u'INSERT INTO treestatus_log (tree, `when`, who, status, reason, tags) VALUES (%s, %s, %s, %s, %s, %s)'] [parameters: (u'tree10', datetime.datetime(2015, 8, 28, 19, 51, 8, 81088, tzinfo=<UTC>), 'human:dmitchell@mozilla.com', u'open', u'', '["backlog"]')]`

djmitche:
Again, this is working fine on my system.  I wonder if [bug 1199301](https://bugzilla.mozilla.org/show_bug.cgi?id=1199301) (upgrade mysql client to 5.6) will help?

I've verified with a general query log that a 5.6 client + 5.6 server successfully send and receive timestamps with `+00:00` at the end.  I don't know why a 5.1 client would generate such a timestamp and then *itself* choke on it, but that seems to be what's happening.

djmitche:
If not it will probably tip me over to using postgres fro RelengAPI.  MySQL is just such crap :(

djmitche:
The other question here is, we've got other code inserting timestamps -- why does *this* query fail?

djmitche:
[...]

PR #231, opened 2015-05-08
Revert setting MySQL timezone to UTC

djmitche:
See #31, #229.


mrrrgn:
Straight forward and clear. Ship it! :+1:

PR #227, opened 2015-05-13
When !can_view and !can_issue, tokenauth view should show *something*

djmitche:
As it is, if you can't issue or view tokens, you get a big 'ol blank screen, which leads to lots of "[what browser version are you using](https://bugzilla.mozilla.org/show_bug.cgi?id=1162556)" kind of debugging.


djmitche:
@mrrrgn how's this look?

mrrrgn:
I like, +:1 after travis is green.

PR #279, opened 2015-06-09
Bug 1154795 - add an endpoint to relengapi that creates and returns a Mozharness archive via S3

lundjordan:
this consolidates my relengapi repo into the now unified build-relengapi and addresses feedback comments regarding:
- hardcoded repo and subdir: now supports any hg repo, branch, and subdir configuration
- hardcoded regions: regions are in the cfg
- blueprints no longer copy skeleton repo style of blueprints/name_of_blueprint.py and instead uses blueprints/name_of_blueprint/{__init__.py, etc}
- my settings.py cfg uses https not http
- I ended up downloading the src archive to disk but in a tempfile

from feedback that is not implemented:
- I did not end up using celery subtasks as I felt it may be overkill for the situation
- I also did not use celery's retry because in the event of a FAILURE, the current logic has it that if the s3 key being requested does not exist, the task to download and upload it will just be re-created again anyway. The need to support retry didn't seem overly important due to this.


lundjordan:
hrm, guess I need to write some docs and tests first :)

will squash and update again

lundjordan:
hrm, validate (which runs coverage) passes for me locally. I'm not sure if a -0.01% decrease in coverage should be held against me too harshly :)

latest pull request contains docs, tests, and pep fixes that was failing in previous PR

djmitche:
True, but we do want to make sure that new code is 100% covered, among other things in https://github.com/mozilla/build-relengapi/blob/master/relengapi/docs/development/contributing.rst

The rationale is that such testing is basically the only thing that could prevent a deploy of relengapi from breaking this service in production -- with continuous deployment, we won't have the luxury of manually clicking around on things that we feel our push may have broken.  So if there's code not covered by tests, then that code could be silently broken by some unrelated refactor, and we wouldn't know until the trees closed.

In this case, it looks like `task_status` is untested, as is the retry handling for the celery task.

djmitche:
Having a deeper review look at this now..

relengapi/docs/deployment/archiver.rst
```diff
@@ -0,0 +1,31 @@
...
+==================
+
+Archiver requires a separate configuration for each endpoint. Within these endpoint configs, you need to state the s3
+buckets, where each bucket represents a region and the template url that will be formatted by query params passed to
+the endpoint.
+
+For AWS credentials, each bucket should be limited to the AWS IAM role corresponding to the AWS credentials. Buckets in
+the configuration are required to be pre-existing.
+
+Finally, Archiver avails of Celery. You will need to provide a broker and back-end.
```
djmitche: s/avails of/uses/

relengapi/docs/deployment/archiver.rst
```diff
@@ -0,0 +1,31 @@
...
+        'secret_access_key': 'secretAccessKeyExample',
+    }
+
+    # for the mozharness endpoint
+    SUBREPO_MOZHARNESS_CFG = {
+        'S3_BUCKETS': [
+            {'REGION': 'us-east-1', 'NAME': 'example-bucket-name-for-us-east-1'},
+            {'REGION': 'us-west-2', 'NAME': 'example-bucket-name-for-us-west-2'}
+        ],
+        "URL_SRC_TEMPLATE": "https://hg.mozilla.org/{repo}/archive/{rev}.{suffix}/testing/mozharness"
```
djmitche: the keys within dictionaries in the config generally aren't capitalized (e.g., `AWS[..]` above)

djmitche: You may also want to note what `{..}` can be interpolated into the template

relengapi/blueprints/archiver/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+    s3 location. If the key does not exist, download the archive from src url, upload it to and
+    return all s3 url locations.
+
+     When the key does not exist, the remaining work will be assigned to a celery background task
+    with a url location returned immediately for obtaining task state updates.
+    """
+    bucket_region = None
+    bucket_name = None
+    for bucket in cfg['S3_BUCKETS']:
+        if region in bucket['REGION']:
```
djmitche: `bucket['REGION']` is a string, not a list -- should this be `==` instead of `in`?

relengapi/blueprints/archiver/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+    else:
+        # something went wrong
+        response['status'] = str(task.info)  # this is the exception raised
+
+    return MozharnessArchiveTask(**response)
+
+
+@bp.route('/mozharness/<rev>')
+@api.apimethod(None, unicode, unicode, unicode, unicode, status_code=302)
+def get_mozharness_archive(rev, repo="mozilla-central", region='us-west-2', suffix='tar.gz'):
```
djmitche: It doesn't seem like `repo` and `region` should have defaults.  If anything, `region` should default to a random selection of region (which is what tooltool does - I'd like to improve that to guess region based on the source IP, but I can upgrade archiver at the same time).  A default `suffix` makes a lot of sense, but should be documented.  Speaking of which, this endpoint function needs a dockstring :)

relengapi/blueprints/archiver/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+
+     When the key does not exist, the remaining work will be assigned to a celery background task
+    with a url location returned immediately for obtaining task state updates.
+    """
+    bucket_region = None
+    bucket_name = None
+    for bucket in cfg['S3_BUCKETS']:
+        if region in bucket['REGION']:
+            bucket_region = bucket['REGION']
+            bucket_name = bucket['NAME']
```
djmitche: It looks like this is just a complicated way of writing `cfg['S3_BUCKETS'][region]`, except that `cfg['S3_BUCKETS']` is a list instead of a dictionary.  Any particular reason to not just make it a dictionary keyed by region?

relengapi/docs/deployment/archiver.rst
```diff
@@ -0,0 +1,31 @@
...
+    CELERY_BROKER_URL='amqp://guest@localhost//'
+    CELERY_BACKEND='amqp'
+
+    AWS = {
+        'access_key_id': 'accessKeyExample',
+        'secret_access_key': 'secretAccessKeyExample',
+    }
+
+    # for the mozharness endpoint
+    SUBREPO_MOZHARNESS_CFG = {
```
djmitche: This should probably be renamed, maybe to `ARCHIVER`.  Alternately, you could split it into `ARCHIVER_BUCKETS` and `ARCHIVER_URL_SRC_TEMPLATE`.

relengapi/blueprints/archiver/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # sanity check
+    if not bucket_name or not bucket_region:
+        valid_regions = str([bucket['REGION'] for bucket in cfg['S3_BUCKETS']])
+        log.warning('Unsupported region given: "{}" Valid Regions "{}"'.format(region,
+                                                                               valid_regions))
+        raise NotFound
```
djmitche: The `NotFound` should have some helpful information, too -- at least "Unsupported region given".

relengapi/blueprints/archiver/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+
+    s3 = current_app.aws.connect_to('s3', bucket_region)
+    bucket = s3.get_bucket(bucket_name)
+    key = '{repo}-{rev}.{suffix}'.format(repo=os.path.basename(repo), rev=rev, suffix=suffix)
```
djmitche: This could lead to collisions if we have repos with the same basename.  Might as well use the whole repo string.

relengapi/blueprints/archiver/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # first, see if the key exists
+    if not bucket.get_key(key):
+        task_id = rev
```
djmitche: If there's a thundering herd of requests for the same revision, all of the Celery tasks will have the same ID -- probably causing an error.  Maybe just use `uuid.uuid4()` here?

djmitche: Oh, I see -- that's the intent, and all requests wait on the same task -- neat!

relengapi/docs/deployment/archiver.rst
```diff
@@ -0,0 +1,31 @@
+Deploying Archiver
+==================
+
+Archiver requires a separate configuration for each endpoint. Within these endpoint configs, you need to state the s3
+buckets, where each bucket represents a region and the template url that will be formatted by query params passed to
+the endpoint.
+
+For AWS credentials, each bucket should be limited to the AWS IAM role corresponding to the AWS credentials. Buckets in
+the configuration are required to be pre-existing.
```
djmitche: By the way, relengapi's AWS IAM role is defined in
  https://github.com/mozilla/build-cloud-tools/blob/master/configs/cloudformation/iam_relengapi.py
and the buckets for tooltool are defined in
  https://github.com/mozilla/build-cloud-tools/blob/master/configs/cloudformation/tooltool.py

relengapi/blueprints/archiver/tasks.py
```diff
@@ -0,0 +1,78 @@
...
+
+def upload_url_archive_to_s3(key, url, region, bucket, suffix):
+    """
+    Given a src url, upload contents to an s3 bucket by a given key.
+    """
+    s3 = current_app.aws.connect_to('s3', region)
+    k = Key(s3.get_bucket(bucket))
+    k.key = key
+
+    # rather than worrying about pointers and seeking, let's avail of a named temp file that is
```
djmitche: "avail" takes an indirect object -- "..avail ourselves of.."  Or maybe just "use" :)

relengapi/blueprints/archiver/tasks.py
```diff
@@ -0,0 +1,78 @@
...
+    # allowed to persist after the file is closed. Finally, when are finished, we can clean up
+    # the temp file
+    resp = requests.get(url)
+    temp_file = tempfile.NamedTemporaryFile(mode="wb", suffix=".{}".format(suffix), delete=False)
+    with open(temp_file.name, "wb") as tmpf:
+        tmpf.write(resp.content)
+    k.set_contents_from_filename(temp_file.name)
+    os.unlink(temp_file.name)  # clean up tmp file
```
djmitche: This unlink should be done in a try/finally block, at least, for when the upload fails.

When we talked earlier about streaming the download directly to S3, I forgot there were multiple regions to upload to.  Streaming to multiple destinations is quite a bit harder, and probably not worth the effort over using a temporary file.  However, as this is written, it performs the download from hg.m.o once for each region, which seems wasteful when the file is sitting right there on disk.  So refactoring this to download to a temporary file and then upload from that file to all regions would be nice.

Using `tempfile.TemporaryFile` would eliminate any worries about unlinking the file.  IIRC getting `set_contents_from_stream` to work was difficult, but `set_contents_from_file` worked OK.

Finally, as with the archiver client, it'd be great to handle the copy from `resp` to the temporary file in a block-by-block manner, and `shutil.copyfileobj` should be able to help with that -- something like `shutil.copyfileobj(resp.raw, tmpf)`.

relengapi/blueprints/archiver/tasks.py
```diff
@@ -0,0 +1,78 @@
...
+                s3_urls[bucket['REGION']] = upload_url_archive_to_s3(key, src_url, bucket['REGION'],
+                                                                     bucket['NAME'], suffix)
+        except Exception as exc:
+            # set a jitter enabled delay
+            # where an aggressive delay would result in: 7s, 49s, and 343s
+            # and a gentle delay would result in: 4s, 16s, and 64s
+            delay = randint(4, 7) ** (current.request.retries + 1)  # retries == 0 on first attempt
+            current.retry(exc=exc, countdown=delay)
+        if not any(s3_urls.values()):
+            status = "Could not upload any archives to s3. Check logs for errors."
```
djmitche: I'm not sure how this would be executed -- `upload_url_archive_to_s3` never returns `None`, so none of the values of this dict should be false whether or not an error occurred.

relengapi/docs/deployment/archiver.rst
djmitche: And one last thing -- this particular configuration design can only ever return one subdirectory, but from any repo on the server.  Why not make the directory (`testing/mozharness`) user-selectable, too?

In fact, you could make the endpoint route match the hg route exactly: `@bp.route('/<repo:path>/archive/<rev_and_suffix>/<subdir:path>')`.  Actually, I'm not sure if the Flask route engine is smart enough to do that, and in fact it doesn't matter -- there's no harm in caching any path like `https://hg.mozilla.org/build/tools/file/default/buildfarm/maintenance/production-masters.json` even though it's not an archive.  From the user's perspective, this is a generic cache-a-thing-from-hg service.

relengapi/blueprints/archiver/tasks.py
```diff
@@ -0,0 +1,78 @@
...
+    k.key = key
+
+    # rather than worrying about pointers and seeking, let's avail of a named temp file that is
+    # allowed to persist after the file is closed. Finally, when are finished, we can clean up
+    # the temp file
+    resp = requests.get(url)
+    temp_file = tempfile.NamedTemporaryFile(mode="wb", suffix=".{}".format(suffix), delete=False)
+    with open(temp_file.name, "wb") as tmpf:
+        tmpf.write(resp.content)
+    k.set_contents_from_filename(temp_file.name)
```
djmitche: It's probably worthwhile setting the content-type here, too.  Although the client you've written doesn't pay attention to the header, other users of this service might, and it'll be easier to have it set from day one.

relengapi/blueprints/archiver/test_tasks.py
```diff
@@ -0,0 +1,91 @@
...
+@moto.mock_s3
+@test_context
+def test_invalid_hg_url(app, client):
+    setup_buckets(app, cfg['SUBREPO_MOZHARNESS_CFG'])
+    rev, repo, suffix, key, = 'foo', 'mozilla-central', 'tar.gz', 'mozilla-central-foo.tar.gz'
+    config = cfg['SUBREPO_MOZHARNESS_CFG']
+    with app.app_context():
+        task = create_and_upload_archive.apply_async(args=[config, rev, repo, suffix, key],
+                                                     task_id='9ebd530c5843')
+    assert "Url not found." in task.info.get('status', {}), "invalid hg url was not caught!"
```
djmitche: Tests shouldn't hit external resources like https://hg.mozilla.org.  Instead, mock out requests to fake the request.  Same for `test_s3_urls_exist_for_each_region` and `test_hg_and_s3_archives_match`.

relengapi/blueprints/archiver/test_tasks.py
```diff
@@ -0,0 +1,91 @@
...
+    setup_buckets(app, cfg['SUBREPO_MOZHARNESS_CFG'])
+    bucket = cfg['SUBREPO_MOZHARNESS_CFG']["S3_BUCKETS"][0]
+
+    src_url = "http://hg.mozilla.org/mozilla-central/archive/203e1025a826.tar.gz/testing/mozharness"
+    with app.app_context():
+        s3_url = upload_url_archive_to_s3(key="203e1025a826", url=src_url, region=bucket["REGION"],
+                                          bucket=bucket["NAME"], suffix='tar.gz')
+
+    src_resp = requests.get(src_url)
+    s3_resp = requests.get(s3_url)
```
djmitche: Huh, I wouldn't have thought this would work with moto, but apparently moto mocks things out in such a way that any HTTP request to AWS -- whether via boto or something else -- is intercepted.  Neat!

relengapi/blueprints/archiver/test_tasks.py
```diff
@@ -0,0 +1,91 @@
...
+    src_file = tempfile.NamedTemporaryFile(mode="wb")
+    s3_file = tempfile.NamedTemporaryFile(mode="wb")
+    with open(src_file.name, "wb") as srcf:
+        srcf.write(src_resp.content)
+        with open(s3_file.name, "wb") as s3f:
+            s3f.write(s3_resp.content)
+            assert cmp(srcf, s3f), "s3 archive based on hg archive does not match!"
```
djmitche: ?? `cmp` doesn't take files as arguments.  Luckily, it returns 0 when its arguments are equal, since two different files are never equal.  So this test passes, but you're asserting that the file handles you wrote to (and not the files' content) are different (rather than the same).

relengapi/blueprints/archiver/test_archiver.py
```diff
@@ -0,0 +1,75 @@
...
+        k.key = key
+        k.set_contents_from_string("Help, I'm trapped in an alternate s3 dimension.")
+
+
+@moto.mock_s3
+@test_context
+def test_accepted_response_when_missing_s3_key(app, client):
+    setup_buckets(app, cfg['SUBREPO_MOZHARNESS_CFG'])
+    resp = client.get('/archiver/mozharness/9ebd530c5843?repo=mozilla-central&region=us-east-1')
+    eq_(resp.status_code, 202, resp.status)
```
djmitche: This hits hg too, doesn't it.. it just doesn't fail because the test doesn't poll the task status..

relengapi/blueprints/archiver/test_archiver.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@moto.mock_s3
+@test_context
+def test_redirect_response_when_found_s3_key(app, client):
+    setup_buckets(app, cfg['SUBREPO_MOZHARNESS_CFG'])
+    create_s3_items(app, cfg['SUBREPO_MOZHARNESS_CFG'], key='mozilla-central-9ebd530c5843.tar.gz')
+
+    resp = client.get('/archiver/mozharness/9ebd530c5843?repo=mozilla-central&region=us-east-1')
+    eq_(resp.status_code, 302, resp.status)
```
djmitche: This I like :)

relengapi/blueprints/archiver/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+    response = {
+        'state': task.state,
+        'src_url': task_info.get('src_url', ''),
+        's3_urls': task_info.get('s3_urls', {})
+    }
+    if task.state != 'FAILURE':
+        response['status'] = task_info.get('status', 'no status available at this point.')
+    else:
+        # something went wrong
+        response['status'] = str(task.info)  # this is the exception raised
```
djmitche: I noticed in playing around with this that `task.info` is not a dictionary in this case, which means that the `task_info.get` calls above will raise an exception before the status can be set correctly.

djmitche:
Sorry, that got really long.  This is in good shape, and most of these comments are relatively minor changes.  I think the only controversial idea is to just accept any old path, tack it onto `https://hg.mozilla.org/`, download it, cache it, and 302 to it.

I don't remember if we talked about how these files will be deleted from S3 -- a lifetime configuration on the S3 bucket?  If that's the case, are you worried about the race condition where S3 deletes a key in between its existence being checked and the client following the 302 redirect?

relengapi/docs/deployment/archiver.rst
```diff
@@ -0,0 +1,29 @@
...
+    CELERY_BROKER_URL='amqp://guest@localhost//'
+    CELERY_BACKEND='amqp'
+
+    AWS = {
+        'access_key_id': 'accessKeyExample',
+        'secret_access_key': 'secretAccessKeyExample',
+    }
+
+    ARCHIVER_S3_BUCKETS = {
+        'us-east-1', 'archiver-us-east-1',
```
djmitche: Change `,` to `:`

relengapi/blueprints/archiver/tasks.py
```diff
@@ -0,0 +1,81 @@
...
+    s3_urls = {}
+
+    # make the source request
+    resp = requests.get(url, stream=True)
+
+    # create a temporary file for it
+    tempf = tempfile.TemporaryFile()
+    # copy the data, block-by-block, into that file
+    resp.raw.decode_content = True
+    shutil.copyfileobj(resp.raw, tempf)
```
djmitche: woo!!

relengapi/blueprints/archiver/tasks.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # write it out to S3
+    for region in buckets:
+        s3 = current_app.aws.connect_to('s3', region)
+        k = Key(s3.get_bucket(buckets[region]))
+        k.key = key
+        k.set_metadata('Content-Type', resp.headers['Content-Type'])
+        # give it the same attachment filename
+        k.set_metadata('Content-Disposition', resp.headers['Content-Disposition'])
+        k.set_contents_from_file(tempf, rewind=True)   # rewind points tempf back to start for us
```
djmitche: ++!

djmitche:
Two one-character changes in the docs and this is good to go :)

PR #283, opened 2015-06-16
Bug 1173648: fix display, cleanup of badpenny tasks

djmitche:
The existing code assumed that tasks were returned in the order they were created, and the tests did nothing to dissuade them (one of the downsides of testing with SQLite).  I bulked up the test a little bit, and refactored the filtering to go job-by-job rather than loading everything into a list, since 7 days of once-a-minute jobs could start to be a lot of RAM.

### DIFF
--- a/relengapi/blueprints/badpenny/cleanup.py
+++ b/relengapi/blueprints/badpenny/cleanup.py
@@ -17,20 +17,29 @@ logger = logging.getLogger(__name__)
 def cleanup_old_jobs(job_status):
     session = current_app.db.session('relengapi')
     Task = tables.BadpennyTask
+    Job = tables.BadpennyJob
 
     old_job_days = current_app.config.get('BADPENNY_OLD_JOB_DAYS', 7)
     old = time.now() - datetime.timedelta(days=old_job_days)
     deleted = 0
 
     for task in Task.query.all():
-        # consider all but the most recent job
-        jobs = reversed(task.jobs[1:])
-        for job in jobs:
-            if job.created_at < old:
-                for log in job.logs:
+        # Iterate until we find a job that's not too old.  Only
+        # delete on the next iteration to avoid deleting the most
+        # recent job.
+        to_delete = None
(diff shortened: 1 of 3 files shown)


relengapi/blueprints/badpenny/cleanup.py
```diff
@@ -17,21 +17,30 @@
...
-        # consider all but the most recent job
-        jobs = reversed(task.jobs[1:])
-        for job in jobs:
-            if job.created_at < old:
-                for log in job.logs:
+        # Iterate until we find a job that's not too old.  Only
+        # delete on the next iteration to avoid deleting the most
+        # recent job.
+        to_delete = None
+        for job in Job.query.filter(Job.task_id == task.id).order_by(Job.created_at):
```
Callek: I personally like explicit asc() or desc() [primarily because I always forget which is default] But I won't block on that

relengapi/blueprints/badpenny/cleanup.py
```diff
@@ -17,21 +17,30 @@
...
                     session.delete(log)
-                session.delete(job)
+                session.delete(to_delete)
+                to_delete = None
                 deleted += 1
 
+            if job.created_at < old:
+                to_delete = job
+            else:
+                break
```
Callek: shouldn't the ``if to_delete`` block be under this block, since otherwise we won't ever delete the last job returned, even if it is the oldest?

Callek:
:+1:

relengapi/blueprints/badpenny/cleanup.py
djmitche: That's the intent -- it avoids deleting the most recent job

relengapi/blueprints/badpenny/cleanup.py
djmitche: Ascending is default.  Just like if I asked you to sort a pile of books by author, you'd start with A and end with Z.

PR #287, opened 2015-06-23
Verify visibilities are set for uploads

djmitche:
This avoids some ISEs on manifests that lack visibilities

### DIFF
--- a/relengapi/blueprints/tooltool/__init__.py
+++ b/relengapi/blueprints/tooltool/__init__.py
@@ -134,10 +134,11 @@ def upload_batch(region=None, body=None):
     except AttributeError:
         raise BadRequest("Could not determine authenticated username")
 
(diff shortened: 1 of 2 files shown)


Callek:
:-1: -- Needs coverage for https://coveralls.io/builds/2883764/source?filename=relengapi%2Fblueprints%2Ftooltool%2F__init__.py#L139 (which I at-a-glance suspect is failing to cover because of a Truthiness Mistake)

djmitche:
WSME is actually verifying that every file has a visibility set.  So there's no way for that code to be executed!  Which leads me to wonder what went wrong in this bug!

djmitche:
Updated, just by removing the unreachable code (I left the test in)

Callek:
:+1:

Issue #236, opened 2015-05-29
Invalid bearer token is treated as no authentication


djmitche:
This is allowing build jobs to succeed even with bogus data in `/builds/relengapi.tok`, as long as they only need publicly available data.  This is actually helpful to @mrrrgn since there's no way to provide relengapi credentials to a try job right now, but otherwise feels like a bug that's hiding an error condition.

djmitche:
Now that https://reviewboard.mozilla.org/r/12485/ is landed, if `$RELENGAPI_TOKEN` isn't set in a TC job, then tooltool.py won't get --authorization-file, and won't send a header, and all is well as far as @mrrrgn's builds.

PR #288, opened 2015-06-23
If an authorization header is present, use it or fail

djmitche:
Failing instead of silently succeeding gives users a better indication of an issue with their credentials.

This will break @mrrrgn's try runs which use "TODO" as a token.  That will need some modification in mozharness, etc. before this lands.


Callek:
defer to @mrrrgn because of the Try run comment here.

djmitche:
@mrrrgn - r?

This fixes #236

djmitche:
Morgan r+'d in irc

PR #293, opened 2015-07-01
bumps relengapi to 3.1.0 for new archiver endpoint

lundjordan:


lundjordan:
I don't think this requires a review so I will merge once travis finishes. Apologies if I am doing this incorrectly

PR #295, opened 2015-07-14
[cosmetic] Disable download link in tooltool if there are no active file instances

acmiyaguchi:
The tooltool page exposes the download link to files with nonexistent instances. This pr disables the download link link so you aren't redirected to a broken link, which  could be the result of removed instances or incomplete file uploads.

### DIFF
--- a/relengapi/blueprints/tooltool/tables.py
+++ b/relengapi/blueprints/tooltool/tables.py
@@ -35,7 +35,8 @@ class File(db.declarative_base('relengapi')):
             size=self.size,
             digest=self.sha512,
             algorithm='sha512',
-            visibility=self.visibility)
+            visibility=self.visibility,
+            has_instances=any(self.instances))
         if include_instances:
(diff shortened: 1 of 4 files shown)


relengapi/blueprints/tooltool/tables.py
```diff
@@ -35,7 +35,8 @@ def to_json(self, include_instances=False):
             size=self.size,
             digest=self.sha512,
             algorithm='sha512',
-            visibility=self.visibility)
+            visibility=self.visibility,
+            download_link=any(self.instances))
```
djmitche: This is a view-specific name, but it's in the data model.  How about `has_instances`?

acmiyaguchi:
I updated the variable to reflect the data better (from `download _link` to `has_instances`).

djmitche:
I don't see anything missing from https://api.pub.build.mozilla.org/docs/development/contributing/#patch-review-checklist so, good to go!

PR #296, opened 2015-07-15
Archiver

lundjordan:
This  combats against creating multiple celery tasks for the same job and manages clean up. It does this by:

* setting a time limit of how long a task can run for (protecting from getting hung tasks)
* setting an expiry of how long a task can be un-actioned in the message queue
* tracking the state of celery tasks with a db backend.
 * Prior to this I only tracked if the state given by celery was 'STARTED'. But celery tasks could also be PENDING or RETRY. You can't trust PENDING as that's the default state for tasks that were never even created.
* explicitly capping the revision at 12 chars.
    * If a buildbot job or user hits the archiver api with both the short hash and the long hash, archiver used to think this was two separate archives. This could be an issue if a bunch of buildbot jobs have the short hash stored in buildbot properties.
* some house keeping with the trackers in the db. If the task goes past it's time limit or is in a finished state and the tracker still exists in the db, a badpenny job will clean it up.
* add more logging to help track health of archiver

Though *much* more work than simply:  `if create_and_upload_archive.AsyncResult(task_id).state != 'STARTED'`, managing the tasks through a tracker should provide us with much more control.

### DIFF
--- a/relengapi/blueprints/archiver/__init__.py
+++ b/relengapi/blueprints/archiver/__init__.py
@@ -14,7 +14,7 @@ from flask import current_app
 from flask import redirect
 from flask import url_for
 
-from relengapi.lib import time
+from relengapi.lib.time import now
 from relengapi.lib import badpenny
 from relengapi.blueprints.archiver import tables
 from relengapi.blueprints.archiver.tasks import create_and_upload_archive, TASK_TIME_OUT
@@ -48,14 +48,17 @@ def update_tracker_state(tracker, state):
         session.rollback()
 
 
-@badpenny.periodic_task(seconds=3600)
+@badpenny.periodic_task(seconds=TASK_TIME_OUT)
 def cleanup_old_tasks():
     """delete any tracker task if it is older than the time a task can live for."""
-    expires_at = time.now() - datetime.timedelta(seconds=TASK_TIME_OUT)
+    session = current_app.db.session('relengapi')
+    expiry_cutoff = now() - datetime.timedelta(seconds=TASK_TIME_OUT)
     table = tables.ArchiverTask
-    for tracker in table.query.query.all().order_by(table.created_at):
-        if tracker.created_at > expires_at:
+    for tracker in session.query(table).order_by(table.created_at):
+        if tracker.created_at < expiry_cutoff:
             delete_tracker(tracker)
+        else:
+            break
 
 
 @bp.route('/status/<task_id>')
@@ -75,13 +78,14 @@ def task_status(task_id):
     """
     task = create_and_upload_archive.AsyncResult(task_id)
     task_tracker = tables.ArchiverTask.query.filter(tables.ArchiverTask.task_id == task_id).first()
+    log.info("checking status of task id {}: current state {}".format(task_id, task.state))
 
     # archiver does not create any custom states, so we can assume to have only the defaults:
     # http://docs.celeryproject.org/en/latest/userguide/tasks.html#task-states
     # therefore, delete our state_id tracker from the db if the celery state is in a finite state:
     # e.g. not RETRY, STARTED, or PENDING
     if task_tracker:
-        if task_tracker.state in FINISHED_STATES:
+        if task.state in FINISHED_STATES:
             delete_tracker(task_tracker)
         elif task_tracker.state != task.state:
             update_tracker_state(task_tracker, task.state)
@@ -115,6 +119,8 @@ def get_hgmo_archive(repo, rev, subdir=None, suffix='tar.gz', preferred_region=N
     :param suffix: the archive extension type. defaulted to tar.gz
     :param preferred_region: the preferred s3 region to use
     """
+    # allow for the short hash and full hash to be passed
+    rev = rev[0:12]
     src_url = current_app.config['ARCHIVER_HGMO_URL_TEMPLATE'].format(
         repo=repo, rev=rev, suffix=suffix, subdir=subdir or ''
     )
@@ -155,6 +161,8 @@ def get_archive(src_url, key, preferred_region):
     # first, see if the key exists
     if not s3.get_bucket(bucket).get_key(key):
         task_id = key.replace('/', '_')  # keep things simple and avoid slashes in task url
+        # can't use unique support: https://api.pub.build.mozilla.org/docs/development/databases/#unique-row-support-get-or-create
+        # because we want to know when the table doesn't exist before creating it
         task_tracker = tables.ArchiverTask.query.filter(tables.ArchiverTask.task_id == task_id).first()
         if task_tracker and task_tracker.state in FINISHED_STATES:
             log.info('Task tracker: {} exists but finished with state: '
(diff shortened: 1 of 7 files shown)


relengapi/blueprints/archiver/__init__.py
```diff
@@ -65,6 +121,8 @@ def get_hgmo_archive(repo, rev, subdir=None, suffix='tar.gz', preferred_region=N
     :param suffix: the archive extension type. defaulted to tar.gz
     :param preferred_region: the preferred s3 region to use
     """
+    # allow for the short hash and full hash to be passed
+    rev = rev[0:12]
```
djmitche: It seems like this should be a client-side requirement, rather than silently discarding parts of the request.  Can you land this as-is, but also update the client to always truncate revs to 12 bytes?  Once that'd deployed, we can update this service to fail with 400 if `len(rev) != 12`.

relengapi/blueprints/archiver/__init__.py
```diff
@@ -36,6 +79,19 @@ def task_status(task_id):
     If state is SUCCESS, it is safe to check response['s3_urls'] for the archives submitted to s3
     """
     task = create_and_upload_archive.AsyncResult(task_id)
+    task_tracker = tables.ArchiverTask.query.filter(tables.ArchiverTask.task_id == task_id).first()
+    log.info("checking status of task id {}: current state {}".format(task_id, task.state))
+
+    # archiver does not create any custom states, so we can assume to have only the defaults:
+    # http://docs.celeryproject.org/en/latest/userguide/tasks.html#task-states
+    # therefore, delete our state_id tracker from the db if the celery state is in a finite state:
```
djmitche: s/finite/final/g

relengapi/blueprints/archiver/__init__.py
```diff
@@ -100,12 +158,30 @@ def get_archive(src_url, key, preferred_region):
...
+    session = current_app.db.session('relengapi')
 
     # first, see if the key exists
     if not s3.get_bucket(bucket).get_key(key):
         task_id = key.replace('/', '_')  # keep things simple and avoid slashes in task url
-        if create_and_upload_archive.AsyncResult(task_id).state != 'STARTED':
-            # task is currently not in progress so start one.
+        # can't use unique support:
+        # api.pub.build.mozilla.org/docs/development/databases/#unique-row-support-get-or-create
+        # because we want to know when the table doesn't exist before creating it
```
djmitche: s/table/row/g

relengapi/blueprints/archiver/__init__.py
```diff
@@ -100,12 +158,30 @@ def get_archive(src_url, key, preferred_region):
...
+        tracker = tables.ArchiverTask.query.filter(tables.ArchiverTask.task_id == task_id).first()
+        if tracker and tracker.state in FINISHED_STATES:
+            log.info('Task tracker: {} exists but finished with state: '
+                     '{}'.format(task_id, tracker.state))
+            # remove tracker and try celery task again
```
djmitche: Why retry on SUCCESS?

djmitche: Oh, because the key doesn't exist, so it's not really successful

relengapi/blueprints/archiver/tasks.py
```diff
@@ -55,6 +58,10 @@ def create_and_upload_archive(self, src_url, key):
 
     Throughout this process, update the state of the task and finally return the location of the
     s3 urls if successful.
+
+    expires after 30m if the task hasn't been picked up from the message queue
+
+    task is killed if exceeds time_limit of an hour after it has started
```
djmitche: Did you get a chance to verify that the task actually does get killed?

relengapi/blueprints/archiver/test_archiver.py
```diff
@@ -89,7 +108,110 @@ def test_task_status_when_failed(app, client):
...
+@moto.mock_s3
+@test_context
+def test_tracker_added_when_celery_task_is_created(app, client):
+    setup_buckets(app, cfg)
+    with mock.patch("relengapi.blueprints.archiver.tasks.requests.get") as get, \
+            mock.patch("relengapi.blueprints.archiver.tasks.requests.head") as head:
+        # don't actually hit hg.m.o, we just care about starting a subprocess and
+        # returning a 202 accepted
+        get.return_value = fake_200_response()
+        head.return_value = fake_200_response()
```
djmitche: ++

djmitche:
Most of the missing coverage here is catching IntegrityErrors.  It'd be nice to get those tested.  I'm not sure that rolling back and returning success is always the right choice, though -- in many cases I suspect it would be better to allow the IntegrityError to bubble up.

https://coveralls.io/builds/3060791/source?filename=relengapi%2Fblueprints%2Farchiver%2F__init__.py#L171 looks like some important code to test.

relengapi/blueprints/archiver/__init__.py
```diff
@@ -100,12 +158,30 @@ def get_archive(src_url, key, preferred_region):
...
+            tracker = None
+        if not tracker:
+            log.info("Creating new celery task and task tracker for: {}".format(task_id))
+            try:
+                session.add(tables.ArchiverTask(task_id=task_id, created_at=now(),
+                                                src_url=src_url, state="PENDING"))
+                session.commit()
+            except sa.exc.IntegrityError:
+                log.error("Could not add tracker {} to ArchiveTask table".format(task_id))
+                session.rollback()
```
djmitche: This in particular should probably abort the request, rather than creating the celery task

relengapi/blueprints/archiver/__init__.py
```diff
@@ -4,20 +4,63 @@
...
+def cleanup_old_tasks():
+    """delete any tracker task if it is older than the time a task can live for."""
+    session = current_app.db.session('relengapi')
+    expiry_cutoff = now() - datetime.timedelta(seconds=TASK_TIME_OUT)
+    table = tables.ArchiverTask
+    for tracker in session.query(table).order_by(table.created_at):
+        if tracker.created_at < expiry_cutoff:
+            delete_tracker(tracker)
+        else:
+            break
```
djmitche: Does this also need to call `AsyncResult(task_id)` to fetch the result from any tasks where the client stopped polling `/status/<task_id>` before the task finished?

relengapi/blueprints/archiver/__init__.py
lundjordan: right. it's only successful in the sense the task completed. I don't think this could happen without the s3 url being there but just in case

relengapi/blueprints/archiver/__init__.py
lundjordan: my assumption here is that the time_out/expiry vars in the celery task will clean this up if it exists and is not finished. And if it is finished, there is no reason to look it up. What would checking the result here do for us? You mean for logging/debugging?

djmitche: I mean mostly to make sure that we're not leaving un-read results hanging around in RabbitMQ queues, which tends to make the MOC ping us and file bugs.

relengapi/blueprints/archiver/__init__.py
lundjordan: client updated for next patch

relengapi/blueprints/archiver/tasks.py
lundjordan: confirmed against a quick 5 second time_out

> curl -i http://127.0.0.1:8010/archiver/status/mozilla-central-49683d4e9ebd.tar.gz_testing_mozharness
HTTP/1.0 200 OK
Content-Type: application/json
Content-Length: 177
Server: Werkzeug/0.10.4 Python/2.7.6
Date: Wed, 15 Jul 2015 21:06:26 GMT

{
  "result": {
    "s3_urls": {},
    "src_url": "",
    "state": "FAILURE",
    "status": "{u'exc_message': u'TimeLimitExceeded(5,)', u'exc_type': u'TimeLimitExceeded'}"
  }
}%

relengapi/blueprints/archiver/__init__.py
lundjordan: I've created a new test to cover lines in this block here: https://coveralls.io/builds/3060791/source?filename=relengapi%2Fblueprints%2Farchiver%2F__init__.py#L171

I also removed the catch when adding or deleting tracking records. In either case, it would be better to bubble up the error and fail the request. I left the IntegrityError catch on update as rolling back  and continuing wouldn't be the end of the world. the tracker state is more of a debugging tool as we rely of the AsyncResult.state for flow behaviour.

PR #298, opened 2015-07-16
bug 1182532 - log task_id not the sql gen id when deleting archiver tracker tasks

lundjordan:
quick typo fix in logging. forgive me if I should have asked for a review on this before landing


djmitche:
Should be OK.  Are you filling out the release notes for each of these releases, too?

PR #299, opened 2015-07-17
Archiver

lundjordan:
Bug 1182532 - archiver is broken - mozharness can't be deployed on slaves


Callek:
:+1: hope this does it

Issue #25, opened 2014-04-04
Use Alembic for DB migrations


djmitche:
This will allow us to support automatic DB updates.

djmitche:
I've got some work on this in my `issue25` branch, but I'm going to suspend it temporarily until I figure out how this is usually handled in AWS.

djmitche:
Selena suggested https://flask-migrate.readthedocs.org/en/latest/

djmitche:
Flask-Migrate looks cool, but the bulk of what it does is wrap the alembic command line, which we need to do differently (`relengapi db`).  And it would be a little tricky to make it work with multiple DB's.  So no real value from that tool.

Here are my notes from the issue25 branch:

Plan
* each blueprint can have its own alembic environment for its own DB (tools
  provided), with config at relengapi/blueprints/foo/alembic.ini; should use
  a version table named after the blueprint
* the relengapi DB is managed by the relengapi project, but other blueprints
  can define other revisions with their own base (e.g., to add minor tables)
  and can depend on specific revisions of the core DB (assuming proper
  notation in setup.py)
* relengapi createdb will create all DB tables from the model, then stamp
  every alembic config it can find with 'head'
* other migrations are performed with the 'alembic' command directly

To Do
* use MetaData(naming_convention=..)
* set up env.py to make an app context (mostly done?)
* set up autogen?
* use `with op.batch_alter_table("some_table") as batch_op`
* validate.sh should make sure there's only one head
* validate.sh should do a complete upgrade, compare(?) to model, and downgrade

Look Into
* symbolic names for relengapi revisions, tagged to dist versions?
* alembic init --template multidb?
* Need to figure out how to do a full upgrade in tests and compare that to the
  model from createdb

djmitche:
As to "how this is usually handled in AWS", typically a DB schema change is spread over several deployments:
 * deploy code that can use the old or new schema, with no schema change (if necessary)
 * deploy the schema change
 * deploy code that requires the new schema
 * deploy schema cleanups

Adding a column is pretty easy: deploy the new extended schema after verifying the existing code will continue to work (with a server-side column default), then deploy the code that depends on that column, and perhaps deploy another schema change to remove the default.

Removing a column inverts the process: deploy a change to stop using the column, then deploy a schema change to delete the column.

So all in all, it requires an easy CD system with good testing and monitoring and a lot of up-front thought about schema changes.

[...]

PR #300, opened 2015-07-21
Add `relengapi alembic` wrapper for database migrations

acmiyaguchi:
This adds a wrapper based on [Flask-Migrate](https://flask-migrate.readthedocs.org/en/latest/) to the relengapi command line. To use it, run `relengapi alembic DBNAME` with the appropriate command. Basic usage follows Flask-Migration, with the exception of having to specify the database name. For example, to run a migration for table change in the relengapi db, run the following:

```
$ relengapi alembic relengapi migrate
$ relengapi alembic relengapi upgrade
```

`relengapi createdb` will also search for available alembic configurations for databases under `relengapi/alembic`, and stamp them with `head`. 

Note that this omits the initialization of alembic configurations with command line utility, they need to be created by hand.  

This should address issue #25 to some degree, although I haven't looked into adding support for the validation script, or possible changes to the documentation.


relengapi/blueprints/base/__init__.py
```diff
@@ -6,14 +6,21 @@
...
 from flask import Flask
 from flask import current_app
 
+import relengapi
+
 from relengapi.lib import subcommands
 
 bp = Blueprint('base', __name__)
 logger = logging.getLogger(__name__)
+__all__ = ['AlembicSubcommand', ]
```
djmitche: I don't see this class defined here

djmitche: Oh, it's imported, hm

relengapi/blueprints/base/alembic_wrapper.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def run(self, parser, args):
+        self.revision(**vars(args))
+
+    def revision(self, directory=None, message=None, autogenerate=False, sql=False,
+                 head='head', splice=False, branch_label=None, version_path=None,
+                 rev_id=None, **kwargs):
+        """Create a new revision file."""
+        config = _get_config(directory)
+        if alembic_version >= (0, 7, 0):
```
djmitche: 0.7.0 is almost a year old now.  We could just require >= 0.7.0 in setup.py and eliminate this check

relengapi/blueprints/base/alembic_wrapper.py
```diff
@@ -0,0 +1,366 @@
...
+                                  'splice onto'))
+        parser.add_argument('--head', dest='head', default='head',
+                            help=('Specify head revision or <branchname>@head to '
+                                  'base new revision on'))
+        parser.add_argument('--sql', dest='sql', action='store_true', default=False,
+                            help=("Don't emit SQL to database - dump to standard "
+                                  "output instead"))
+        parser.add_argument('--autogenerate', dest='autogenerate',
+                            action='store_true', default=False,
+                            help=('Populate revision script with andidate migration '
```
djmitche: *c*andidate

relengapi/lib/alembic.py
```diff
@@ -0,0 +1,71 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
djmitche: Is this whole file auto-generated?  If so, probably best to indicate that this is covered under Alembic's MIT license, rather than the MPL.  Same for any other autogenerated content.

djmitche:
Those are all *very* minor points.  In general, I think this looks great.

I see the mako templates in here -- what's preventing autogeneration of migration scripts?

Docs on how to use this functionality will be important.  The verification stuff, too -- that's the only means we have of testing that the migration produces the same schema as `metadata.create_all()`.

Finally, we'll need some tests (coveralls says coverage dropped) and release notes.

relengapi/blueprints/base/__init__.py
djmitche: I see this is basically importing `alembic_wrapper` for its side effects.  A minor issue: please use absolute imports (`from relengapi.blueprints.base import alembic_wrapper`).  But a more substantial issue: maybe it's time to get rid of the `base` blueprint, which only defines subcommands at this point (it used to do more).  Each of those subcommands could move somewhere under `relengapi/lib`.  Since you already have `relengapi/lib/alembic.py`, perhaps that's a good place for these subcommands?

If you like this plan, please open another issue to move the remaining subcommands out of the base blueprint.

relengapi/lib/alembic.py
acmiyaguchi: This one is slightly modified from the autogenerated one, with a few changes here and there.

acmiyaguchi:
Nothing stops us from autogenerating migration scripts, that functionality works well enough (downgrading on sqlite3 is an issue I've found, but that's an alembic issue). I took out functionality for initializing a new database, since it takes a little more work than the thin wrapper. 

I'll take a look into it, and docs and verification, now that everything else looks good for the most part.

acmiyaguchi:
@djmitche r? 

I've added support for database initialization (`relengapi alembic DBNAME init`), as well as coverage tests and validation tests. The validation script checks that there is a single head (or none in the case that no migrations have been applied), as well as running all migrations down and up. Validation is done via mysqldump and diffing the original dump with the modified version. I've also added documentation under the deployment folder.

Also, before I write the release notes, would this bump it up to 3.1.4?

djmitche:
I lied about release notes -- for *Buildbot* we need release notes in the PR, but for relengapi they are created from commit messages when we push a release.  Sorry about that.

I'll have a look!

validate.sh
```diff
@@ -113,6 +113,53 @@ head -n2 ${tmpbase}/covreport
...
+for filename in `find relengapi/alembic -type f -name "alembic.ini" ! -path template`; do
+    [ `alembic -c $filename heads | wc -l` -le 1 ] || not_ok "multiple heads exist in migrations"
+done
+finish_step
+
+# run migration tests, only if we're on travis ci
+if $TRAVIS; then
+start_step "check database migrations"
+# set up the settings file
+test_dir=$(mktemp -d)
```
djmitche: need to clean this up

relengapi/docs/deployment/alembic.rst
```diff
@@ -0,0 +1,60 @@
+Database Migrations
+===================
+
+Database migrations are handled through alembic through the command line tool ``relengapi alembic``.
+This tool is built based on Flask-Migrate_, and it's usage is roughly one to one.
```
djmitche: "its"

djmitche:
This looks great.  I want to play with it a bit - add a migration script, run it, etc.  The comments above are pretty minor, just things that I noticed while reading the diff.  The patch coverage is a little light, though -- https://codecov.io/github/mozilla/build-relengapi/compare/master...8354604880ab863baffcb74917ffffda807d7cfe.  It'd be good to get `_get_config` and `Config.get_template_directory` covered, along with the error conditions in `AlembicSubcommand.run`.  The methods called by `run` in each sub-subcommand seem a step too far, though -- they're just passing on function calls.  So I'd be comfortable with marking those as `# pragma: no-cover` if you don't see a better way to get them covered.  The last bit is the offline migration support.  If that works, we should test it.  If it doesn't, we should just remove it.

djmitche:
Well, I'm impressed!  Offline migrations *do* work, and will probably be useful for us in deployment, so they should be tested.

I do think we need some docs for relengapi devs who just want to add a column or make some other fairly simple change.  There are a bunch of gotchas, not least of which is sqlite's poor support for schema modifications.  Since `relengapi createdb` works for sqlite, I think it's safe to say "if you're writing a migration, test it against MySQL."  If you'd like, I can work on the doc - just let me know.

The other thing to worry about is coverage of migration scripts.  Even on Travis, the migrations run outside the `coverage run` invocation in `validate.sh`, so they aren't counted toward the total.  Probably the easiest thing would be to change `validate.sh` to invoke the upgrade and downgrades via `coverage run`.

This is really awesome -- thanks!  And I'm sorry for taking a while to review.

djmitche:
Here's an (untested) sketch of the validate.sh changes for coverage: https://github.com/djmitche/build-relengapi/commit/03790771fcb556c47d31d420dcd5fb63894b3034

acmiyaguchi:
Thanks Dustin, its no problem. It'd be awesome if you could help me with work on the documentation. Since I'm probably not going to be that prolific of a user, it would be nice to get all the parts I would probably gloss over.

And I'm fine with a `#pragma: no-cover` too, its kind of tricky to get the sub-commands running. Is that pragma for the tool that does code coverage locally? I'll get those functions I missed covered in the meanwhile, as well as offline migrations.

djmitche:
Yes, it's an instruction to the coverage tool.  If you grep for it, you can find a few other uses in the codebase.  I'll get a commit up with the docs and link to it here, so you can cherry-pick it into your branch.

djmitche:
Here you go -- https://github.com/djmitche/build-relengapi/commit/0b979142b7c931d6a3444141b47c2b4f5863f531

acmiyaguchi:
r? @djmitche 

Since the last time, I've upped the code coverage. I've also added a `--no-logging` or `-Q` option to the command line tool which makes it easier to read out the standalone output from alembic. I've also added validation tests for offline migrations.

djmitche:
You do beautiful work!!

Callek:
I both love you for this (awesome work that I've been waiting for!) and despise it (more work for me when I write new code). Haha, either way Thanks A Lot for this contribution!

djmitche:
For the kind of backward-compatible schema changes we've been making (adding new tables), it's really easy.

PR #305, opened 2015-07-23
support adding response headers with a decorator

djmitche:
Another useful bit I found while porting treestatus

### DIFF
--- a/relengapi/docs/development/http.rst
+++ b/relengapi/docs/development/http.rst
@@ -4,17 +4,25 @@ HTTP Helpers
 Response Headers
 ----------------
 
-.. py:function:: relengapi.lib.http.response_headers(*headers)
+.. py:function:: relengapi.lib.http.response_headers(*headers, [status_codes=..])
 
     :param headers: sequence of (header, value) tuples
+    :param status_codes: apply only to these response status codes
 
     Returns a decorator for view functions which will add the given headers to non-error responses.
     Insert this decorator just after the ``route`` decorators and before ``apimethod`` or any permission-checking decorators.
 
+    The matching status codes can be specified in a variety of ways:
+
+        * as a function: ``status_codes=lambda c: 200 <= c < 300 or c == 302``
+        * as a single code: ``status_codes=404``
+        * as a class: ``status_codes=3xx``
+        * as an iterable: ``status_codes=[200, 204, 302]``
+
 Use this function as follows::
 
     @bp.route('/some/path')
-    @http.response_headers(('cache-control', 'no-cache'))
+    @http.response_headers(('cache-control', 'no-cache'), status_codes='2xx')
     @p.foo.bar.require()
     def my_view():
         ..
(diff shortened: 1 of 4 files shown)


djmitche:
OK, should be good now :)

relengapi/docs/development/http.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Response Headers
+----------------
+
+.. py:function:: relengapi.lib.http.response_headers(*headers)
+
+    :param headers: sequence of (header, value) tuples
+
+    Returns a decorator for view functions which will add the given headers to non-error responses.
+    Insert this decorator just after the ``route`` decorators and before ``apimethod`` or any permission-checking decorators.
```
Callek: pretty soon we'll need a lexical grammar for decorators of our flask functions, I'm not a fan of arbitrary(ish) rules for decorator order, to be clear.  [not blocking just thinking out loud]

relengapi/lib/http.py
```diff
@@ -0,0 +1,23 @@
...
+def response_headers(*headers):
+    @wrapt.decorator
+    def wrapper(wrapped, instance, args, kwargs):
+        # skip apimethod __data_only__ calls
+        if '_data_only_' in kwargs:
+            return wrapped(*args, **kwargs)
+        # mark the response to avoid caching; insert this decorator just after
+        # the route decorators
+        resp = make_response(wrapped(*args, **kwargs))
+        if 200 <= resp.status_code < 400:
```
Callek: why wouldn't we want to, say, set no-cache on say a 404 if we're explicitly setting this on an api endpoint?

(Mostly curious, again, not-blocking)

relengapi/tests/test_lib_http.py
```diff
@@ -0,0 +1,65 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
Callek: I'd love an additional test to check with permissions (both correct and incorrect) for sanity sake that correct perms yields the header, and incorrect *doesn't* (as thats an error response)

Callek:
:+1: as is, but I do desire the comments answered anyway.  If you add the extra tests I suggest the :+1: can be carried forward

relengapi/docs/development/http.rst
djmitche: Yeah, I think its is a common problem with Flask, which is very decorator-heavy.  However, tests should reveal any errors here.

relengapi/lib/http.py
djmitche: I got this from a Flask snippet, which Firefox 38 has helpfully deleted from my history.  I don't have any particularly good answer here, though.

djmitche: Oh, now I remember -- because this code isn't exclusively for no-cache headers.  For other header types, this behavior makes more sense.  I'll add an optional parameter to specify.

djmitche:
Is this better?  Patch coverage is back up to 100%.

Callek:
:+1:

PR #307, opened 2015-07-23
add api.loads and api.dumps

djmitche:
Regular old `json.dumps` will barf on a WSME type; and `json.loads` will successfully load such a thing, but represent it as a dictionary rather than a WSME object.  These little gems take care of both problems.

### DIFF
--- a/relengapi/docs/development/api_methods.rst
+++ b/relengapi/docs/development/api_methods.rst
@@ -34,10 +34,26 @@ See the WSME_ documentation for more detail.
 
 As a utility, an arbitrary JSON Object can be described with this class:
 
-.. attribute:: relengapi.lib.api.jsonObject
+.. py:attribute:: relengapi.lib.api.jsonObject
 
     A WSME custom type describing an arbitrary JSON object.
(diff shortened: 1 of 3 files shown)


relengapi/docs/development/api_methods.rst
```diff
@@ -34,11 +34,27 @@ See the WSME_ documentation for more detail.
...
+
+        print(api.dumps(JsonThing, my_thing))
+
+.. py:function:: relengapi.lib.api.loads(datatype, obj)
+
+    Load the given object from a JSON string, assuming it is of the given WSME datatype.
+    For example::
+
+        api.loads(JsonThing, data)
+
```
Callek: Can you expend the examples a bit to have "input" and "output" results in the example. Reading this doc without the code (below) is pretty confusing to me what this expects/yields.

Callek: Just putting in-place something akin to what you have in the test should be enough.

Callek:
:+1: with some doc clarity possibly as suggested

djmitche:
Better?

Callek:
:+1:

PR #308, opened 2015-07-23
Port Treestatus

djmitche:
This makes some pretty substantial changes to the sheriff/admin UIs, but is largely the same for users.

Functionally, a big difference here is that trees are no longer removed from outstanding changes if an update is made with the "remember" checkbox unchecked.  That allows the common workflow of closing trees with reason "OMG!", then later following up with a new reason ("bug 1234") without remembering the latter change.  When the first change is undone, the status returns to what it was before.

The DB changes quite a bit, too -- new table names, column names, and column types.  Transitioning from one to the other shouldn't be too hard, though - a decent mysqldump should do the trick.  We'll also need to change the hghook to point at a different URL.  I had a look at the hook already, and it's an easy fix -- just needs to be landed.  During the few minutes of transition, we'd just need to update tree statuses in two places.

### DIFF
--- a/relengapi/blueprints/treestatus/__init__.py
+++ b/relengapi/blueprints/treestatus/__init__.py
@@ -35,8 +35,6 @@ TREE_SUMMARY_LOG_LIMIT = 5
 
 # TODO: replicate cache control headers
 # TODO: use elasticache
-# TODO: rename columns, tables, move to relengapi DB
-# TODO: remove user table, tokens
 # TODO: test deleting a tree with logs or history
 # TODO: add "refresh", run it periodically
 
@@ -61,7 +59,7 @@ def update_tree_status(session, tree, status=None, reason=None,
             tree=tree.tree,
             when=relengapi_time.now(),
             who=str(current_user),
-            action=status,  # action, status same thing
+            status=status,
             reason=reason,
             tags=tags)
         session.add(l)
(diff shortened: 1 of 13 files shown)


djmitche:
@catlee r?

relengapi/blueprints/treestatus/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+def tree_cache_get(tree):
+    with _get_mc() as mc:
+        if not mc:
+            return None
+        data = mc.get(tree.encode('utf-8'))
+        if not data:
+            return
+        print "HIT"
```
catlee: this print statement should probably go away, or be turned into a logging.debug call

relengapi/blueprints/treestatus/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+def index():
+    return angular.template('index.html',
+                            url_for('.static', filename='treestatus.js'),
+                            url_for('.static', filename='treestatus.css'),
+                            stack=api.get_data(get_stack),
+                            trees=api.get_data(get_trees))
+
+
+@bp.route('/details/<path:tree>')
+def show_tree_details(tree):
```
catlee: what is this route for? how is it different than /trees/$tree?

djmitche: `/trees/<path:tree>` is an API endpoint (returns JSON).  This is a UI endpoint.

catlee:
🚢

djmitche:
(rebased before merging)

PR #309, opened 2015-07-27
Archiver

lundjordan:
Bug 1182532 - archiver is broken - mozharness can't be deployed on slaves

https://bugzilla.mozilla.org/show_bug.cgi?id=1182532#c17

### DIFF
--- a/relengapi/blueprints/archiver/__init__.py
+++ b/relengapi/blueprints/archiver/__init__.py
@@ -114,7 +114,7 @@ def task_status(task_id):
             create_and_upload_archive.apply_async(args=[task_tracker.src_url, task_tracker.s3_key],
                                                   task_id=task.id)
             response['state'] = 'RETRY'
-            response['status'] = 'Task has expired from pending too long. Re-creating task.'
+            response['status'] = 'Task has expired from pending for too long. Re-creating task.'
         elif task_tracker.state != task.state:
             update_tracker_state(task_tracker, task.state)
 
--- a/relengapi/blueprints/archiver/__init__.py
+++ b/relengapi/blueprints/archiver/__init__.py
@@ -27,6 +27,7 @@ bp = Blueprint('archiver', __name__)
 log = logging.getLogger(__name__)
 
 GET_EXPIRES_IN = 300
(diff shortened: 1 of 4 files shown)


relengapi/blueprints/archiver/__init__.py
```diff
@@ -103,6 +101,23 @@ def task_status(task_id):
...
+    # archiver does not create any custom states, so we can assume to have only the defaults:
+    # http://docs.celeryproject.org/en/latest/userguide/tasks.html#task-states
+    # therefore, delete our state_id tracker from the db if the celery state is in a final state:
+    # e.g. not RETRY, STARTED, or PENDING
+    if task_tracker:
+        if task.state in FINISHED_STATES:
+            delete_tracker(task_tracker)
+        elif task.state == "PENDING" and task_tracker.pending_expires_at < now():
+            log.info("Task {} has expired from pending too long. Re-creating task".format(task.id))
+            renew_tracker_pending_expiry(task_tracker)  # let exceptions bubble up before moving on
```
djmitche: I'm not clear why this is here -- if it's expired, why renew it?

djmitche:
I'm with callek on hoping this turns out to be unnecessary, but if the logic is fixed I'm happy to see it merged.

relengapi/blueprints/archiver/__init__.py
lundjordan: the idea is that if a task is still pending after what is considered a tolerable amount of time to be in pending, we assume this task is going to be in a perma-PENDING state like I was seeing here: https://bugzilla.mozilla.org/show_bug.cgi?id=1182532#c11

basically, If we get into perma-PENDING, the tracker will continue to exist and lock any task based on the same repo+rev from starting. That is until badpenny comes around and scoops up old trackers (this is why test jobs ended up succeeding with archiver an hour after the builds jobs). Here we are allowing relengapi to create a new celery task if the original never started.

djmitche: Ah, I see, and keeping the same tracker -- hence the renewal of that tracker

PR #312, opened 2015-07-29
Run celery.backend_cleanup via Badpenny

djmitche:
@lundjordan what do you think?

### DIFF
--- a/relengapi/lib/celery.py
+++ b/relengapi/lib/celery.py
@@ -62,9 +62,12 @@ def task(*args, **kwargs):
 @badpenny.periodic_task(3600 * 24)
 def backend_cleanup(js):
     # Celery's built-in backend_cleanup task depends on Beat, which relengapi
-    # does not use, so we just run it via badpenny
-    backend_cleanup_task = current_app.celery.tasks['celery.backend_cleanup']
-    backend_cleanup_task.apply()
+    # does not use, so we just run it via badpenny.  These conditions attempt
+    # to replicate those in Celery's `celery/beat.py`
+    c = current_app.celery
+    if c.conf.CELERY_TASK_RESULT_EXPIRES and not c.backend.supports_autoexpire:
+        backend_cleanup_task = c.tasks['celery.backend_cleanup']
+        backend_cleanup_task.apply()
 
 
 @celeryd_after_setup.connect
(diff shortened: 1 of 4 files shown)


relengapi/lib/celery.py
```diff
@@ -58,6 +59,15 @@ def wrap(fn):
     return inner(**kwargs)
 
 
+@badpenny.periodic_task(3600 * 24)
+def backend_cleanup(js):
+    # Celery's built-in backend_cleanup task depends on Beat, which relengapi
+    # does not use, so we just run it via badpenny
+    backend_cleanup_task = current_app.celery.tasks['celery.backend_cleanup']
+    print "got herez"
```
Callek: debugz

djmitche: gonez, thanks :)

relengapi/lib/celery.py
```diff
@@ -58,6 +59,14 @@ def wrap(fn):
     return inner(**kwargs)
 
 
+@badpenny.periodic_task(3600 * 24)
+def backend_cleanup(js):
+    # Celery's built-in backend_cleanup task depends on Beat, which relengapi
+    # does not use, so we just run it via badpenny
+    backend_cleanup_task = current_app.celery.tasks['celery.backend_cleanup']
+    backend_cleanup_task.apply()
```
lundjordan: looks good.

side-note: celery src suggests this is a builtin like chord but I can't figure out how to import it outside looking it up by name in list of tasks: https://github.com/celery/celery/blob/master/celery/app/builtins.py#L33

lundjordan:
lgtm - I noticed that beat has a check for looking if 1) we actually want tasks to expire (CELERY_TASK_RESULT_EXPIRES != None) and 2) if the backend supports autoexpire (I'm guessing for things like db backends): https://github.com/celery/celery/blob/master/celery/beat.py#L201

Do we want those checks?

I'm worried if we do not, relengapi will trip at this badpenny step for environments where devs don't set a backend, disable CELERY_TASK_RESULT_EXPIRES, or the backend is not a db backend.

djmitche:
That's not a bad point - I could probably duplicate those checks easily.

I think getting the task from the task list is probably more durable than trying to import it.

djmitche:
How's that?

lundjordan:
r+ lgtm +1

PR #315, opened 2015-07-31
Icing on the Alembic Cake

djmitche:
@acmiyaguchi 

This just makes validate a little nicer, and allows me (and anyone with a MySQL instance) to run the MySQL tests locally, rather than waiting for Travis.


djmitche:
I'll fix up the coverage.. oops!

acmiyaguchi:
Nice, everything works locally for me.

PR #316, opened 2015-07-31
[alembic] Apply full downgrade and upgrades to db migrations in validation.sh

acmiyaguchi:
This fixes a bug dealing with offline migration tests. Downgrades offline do not allow for incremental changes using relative values, ie we cant go from `<head>:<head-1>`, and then from `<head-1>:<head-2>`. This will apply a full downgrade and a full upgrade, but through the alembic mechanism rather than from bash.


djmitche:
+1

PR #317, opened 2015-07-31
Bug 1167636 - Add option for TTL

acmiyaguchi:
Now with database migrations out of the way, here is the server implementation of the time to live value. This pull request adds a new column to tooltool.table.File that tracks expiring file instances. A recurring badpenny job will groom files that have expired. 

If there is no TTL field in the upload request, files will be marked to be kept forever. TTL values can be updated if there is a previous TTL value, but not if the file is marked permanent. This means you can extend the life of an expired file by re-uploading it with a new TTL value. 

References:
- [Bug 1167636](https://bugzilla.mozilla.org/show_bug.cgi?id=1167636)
- [[build-tooltool] PR #23](78bc8d0b2502b2e4cb30f9393ba7d1b2fc949d71)


relengapi/alembic/relengapi/versions/4bc24d5bbee6_.py
```diff
@@ -0,0 +1,28 @@
+"""Add column for tracking the time to live value
+Revision ID: 4bc24d5bbee6
```
djmitche: We should make a practice of giving names to the revisions -- `4bc24d5bbee6_add-tooltool-ttl.py` rather than just `4bc24d5bbee6_.py`

relengapi/blueprints/tooltool/grooming.py
```diff
@@ -27,6 +27,19 @@ def check_pending_uploads(job_status):
 
 
 @badpenny.periodic_task(seconds=3600)
```
djmitche: We can probably do this daily rather than hourly

relengapi/blueprints/tooltool/grooming.py
```diff
@@ -101,6 +114,19 @@ def check_file_pending_uploads(sha512):
...
 
+def remove_file_pending_deletion(session, file):
+    """Remove all instances of a file"""
+    config = current_app.config['TOOLTOOL_REGIONS']
+    key_name = util.keyname(file.sha512)
+    for instance in file.instances:
+        target_bucket = config[instance.region]
+        conn = current_app.aws.connect_to('s3', instance.region)
+        bucket = conn.get_bucket(target_bucket)
+        bucket.delete_key(key_name)
```
djmitche: This should probably handle failure (if the key doesn't exist) gracefully.  That will make the cleanup idempotent, even if it fails mid-cleanup (after having deleted the key but before having deleted the DB row).

relengapi/blueprints/tooltool/static/tt-result-file.html
```diff
@@ -15,5 +15,8 @@
         <span class="tt-filesize">{{res.size}} bytes</span> -
         <span ng-if="res.visibility == 'public'" class="text-muted">PUBLIC</span>
         <span ng-if="res.visibility == 'internal'" class="text-danger">INTERNAL</span>
+        <span ng-if="res.expires">
+            - <span class="text-muted">{{ res.expires }}</span>
```
djmitche: Should probably say "expires" in here too :)

relengapi/blueprints/tooltool/tables.py
```diff
@@ -31,12 +33,17 @@ def batches(self):
         return {bf.filename: bf.batch for bf in self._batches}
 
     def to_json(self, include_instances=False):
+        if self.expires is None or self.expires < time.now():
+            expires = ''
+        else:  # pragma: no cover
+            expires = 'Expires ' + self.expires.strftime('%Y-%m-%d %H:%M')
```
djmitche: The JSON should contain the actual datetime, rather than a user-visible message (which clients would then have to parse to get the expiration time)

relengapi/blueprints/tooltool/types.py
```diff
@@ -31,6 +31,12 @@ class File(wsme.types.Base):
     #: Boolean to determine whether the file is available to download.
     has_instances = bool
 
+    #: The time to live for the file, in days
+    ttl = int
+
+    #: String that displays the expiration date
+    expires = unicode
+
```
djmitche: At this point, one of these is for input (ttl, a relative time) and one for output (expires, an absolute time).  It'd be nice to have only one key that works for both input and output.

There's a lot of value to the input being a relative time, so perhaps the output could also be a relative time, just adjusted based on the part of the TTL that has already elapsed.  So if I upload with a TTL of 7, then get the file 2 days later, the returned TTL is 5.

This might be a little easier (avoiding rounding errors) with a one-second resolution instead of one-day.  The `--ttl` option could still take days, since that's the most common use-case for users, and just translate that to seconds before sending to the server.

djmitche:
@acmiyaguchi I'd like to deploy a new relengapi today or early tomorrow, and it'd be great to have this included.  I think it's just a few edits away from finished (as is https://github.com/mozilla/build-tooltool/pull/23).  Do you want to wrap it up?

acmiyaguchi:
I have kept the expires field to the json, but have changed it to be milliseconds (int), relative to epoch. Using ttl as an output field breaks a few tests in a nontrivial way, so its easier to separate them.

I've also made the ttl field represent seconds, rather than days.

relengapi/lib/time.py
```diff
@@ -8,3 +8,8 @@
 
 def now():
     return datetime.datetime.utcnow().replace(tzinfo=pytz.UTC)
+
+
+def seconds_from_epoch(time):
+    return int(
+        (time - datetime.datetime.utcfromtimestamp(0).replace(tzinfo=pytz.UTC)).total_seconds())
```
djmitche: ..or just `time.time()`?

djmitche:
Hm, the last few commits ended up not changing a lot :)

I'd really like to have one field with a consistent meaning that indicates when the file will expire.  Exactly what format it takes isn't too important -- a simple datetime would work, or a time relative to the request -- but it should be consistent, and a single field.

As for trouble with testing, there's a `with set_time(..):` available for testing (grep for it) that can be used to "fix" the current time for clock-sensitive tests.  Does that help with the issues you were seeing?

djmitche:
@acmiyaguchi is this ready for more review?

acmiyaguchi:
It's been a while, but I think I left it in a functional state ready for reviewing.

Callek:
@djmitche just to comment on where it was said:

>>There's a lot of value to the input being a relative time, so perhaps the output could also be a relative time, just adjusted based on the part of the TTL that has already elapsed. So if I upload with a TTL of 7, then get the file 2 days later, the returned TTL is 5.<<

I find API's that return relative time, unless EXPLICITLY a time API that is intentionally returning relative is a bad idea. especially because it makes it harder to work with in many circumstances. is "5" 5 days, 5 hours, 5 minutes or 5 seconds, without having to leave your code and look at docs over again, you don't know.

Also, when was "5" relative to, since the duration of "return value generated" --> "my script is able to process it" could differ, and thus there could be significant (significant based on user-of-api not api-itself) drift in the actual ttl if not an absolute time.

djmitche:
True, there are advantages on either side.  Since this isn't a terribly precise value, I think this solution is fine.  I'm going to push a small fix to actually test the `File.to_json` method and update the comments, BTW.

djmitche:
Aside from the comments, above, Callek has expressed some concern that this feature might unintentionally delete files needed for production/release builds, leaving us with no existing copy of the necessary file.  So I'm backing out in c371e82418a2f6e71d7254c94744ed62493a1b47

Issue #282, opened 2015-06-15
Use `app.db.session` instead of `Session` in mapper tests


djmitche:
We have
```
Session = sessionmaker()


def db_setup(app):
    engine = app.db.engine("mapper")
    Session.configure(bind=engine)
    session = Session()
```
when the same could be accomplished with `app.db.session("mapper")`

PR #318, opened 2015-08-05
Bug 1174215: Add endpoint to return all configured project names

hwine:
https://bugzil.la/1174215 has details

### DIFF
--- a/relengapi/blueprints/mapper/test_mapper.py
+++ b/relengapi/blueprints/mapper/test_mapper.py
@@ -27,29 +27,24 @@ SHAFILE = "%s %s\n%s %s\n%s %s\n" % (
     SHA2, SHA2R,
     SHA3, SHA3R)
 
-Session = sessionmaker()
-
 
 def db_setup(app):
-    engine = app.db.engine("mapper")
-    Session.configure(bind=engine)
-    session = Session()
+    session = app.db.session('mapper')
     project = Project(name='proj')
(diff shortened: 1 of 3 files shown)


relengapi/blueprints/mapper/test_mapper.py
```diff
@@ -45,6 +45,17 @@ def db_teardown(app):
     engine.execute("delete from projects")
```
petemoore: I think we should find out why the ORM version of these two delete SQL statements isn't working, and get that to work. It is maybe stylistic more than anything else - realistically we probably won't switch out the back end to something that wouldn't understand `delete from` statements, but it would be good to keep it clean.

petemoore:
Nice work Hal! Not sure what the error above is though in the automatic checks: `TypeError: Converting circular structure to JSON` - do we need to worry about that?

djmitche:
We're dogfooding Morgan's taskcluster-github, and it's not reporting status correctly -- you can ignore the "default" line.

djmitche:
Looks good!  I added a few fixups for DB use in the tests in https://github.com/djmitche/build-relengapi/commit/eab83ec4aa2b9d27707066f31bf9ced735a0ac3b -- can you cherry-pick that?

djmitche:
(and have a look while you do, I mean)

djmitche:
that will, among other things, fix #282

PR #319, opened 2015-08-05
Bug 1191361 - Trees closed for archiver failures

lundjordan:
What's happening in 1191361 is we are hanging on polling hgmo. This patch adds a timeout to the requests call. It also removes the need to poll hgmo twice (we used head(hgmo_url) and get(hgmo_url) before).

It was mentioned in irc that the overall task timeout should be tweaked as well. Right now, that is set to 1hr which is too generic/generous for mozharness archives. However, changing that is a more invasive change. I say that because, as it is now, we have set the timeout in the task definition (not in the apply_async call). In order to change the overall timeout, we would need to make this configurable and defined at each apply_async call so that we could have a quick timeout for mh archives or a long timeout for larger subdir archives.

What I am trying to do in this patch is at least stop hgmo calls hanging. Configuring the overall task timeout seems like a follow up.

@djmitche what do you think?


lundjordan:
hrm, maybe I should have rebased upstream change in to my development archiver branch. not sure why the above error is happening

relengapi/blueprints/archiver/tasks.py
```diff
@@ -24,32 +24,37 @@
...
+            k.key = key
+            k.set_metadata('Content-Type', resp.headers['Content-Type'])
+            # give it the same attachment filename
+            k.set_metadata('Content-Disposition', resp.headers['Content-Disposition'])
+            k.set_contents_from_file(tempf, rewind=True)   # rewind points tempf back to start
+            s3_urls[region] = s3.generate_url(expires_in=SIGNED_URL_EXPIRY, method='GET',
+                                              bucket=buckets[region], key=key)
+        status = "Task completed! Check 's3_urls' for upload locations."
+    else:
+        status = "Url not found. Does it exist? url: '{}', response: '{}' ".format(url,
```
djmitche: resp != 200 does not imply "Url not found.."  Perhaps "Non-OK response fetching .."?

djmitche:
Loooks good, aside from missing coverage of 12% of the patch :)

relengapi/blueprints/archiver/tasks.py
lundjordan: updated. using requests builtin raise_for_status() now: http://docs.python-requests.org/en/latest/user/quickstart/#response-status-codes

relengapi/blueprints/archiver/tasks.py
```diff
@@ -24,11 +24,18 @@
...
-    logger.info(
-        'Key to be uploaded to S3: %s - downloading and unpacking archive from src_url', key)
-    # make the source request
-    resp = requests.get(url, stream=True)
-
+    logger.info('Key to be uploaded to S3: %s - Verifying src_url: %s', key, url)
+    resp = requests.get(url, stream=True, timeout=60)
+
+    try:
+        resp.raise_for_status()
```
djmitche: Neat!  +1

PR #290, opened 2015-08-20
Make `relengapi.apimethod`, etc. exist when importing `relengapi`

djmitche:
`relengapi` used to be a namespace package, which meant it couldn't have attributes at import time, and they were added only later when `relengapi.app` was imported.

It's no longer a namespace package, so this is no longer required.

### DIFF
--- a/relengapi/app.py
+++ b/relengapi/app.py
@@ -4,7 +4,6 @@
 
 import os
 import pkg_resources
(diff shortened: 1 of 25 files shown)


djmitche:
ready for review

Callek:
:+1:

PR #324, opened 2015-08-26
Add compat endpoints for compatibility with old treestatus

djmitche:
This is part of the deployment strategy: proxy requests from the old treestatus URL to the new.  But RelengAPI wraps every API response in {'result': ..}, while the old treestatus did not.

### DIFF
--- a/relengapi/blueprints/treestatus/__init__.py
+++ b/relengapi/blueprints/treestatus/__init__.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import flask
 import json
 import logging
 import sqlalchemy as sa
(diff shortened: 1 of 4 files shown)


djmitche:
The Travis failures are #325, unrelated to this patch.

relengapi/blueprints/treestatus/__init__.py
```diff
@@ -133,6 +134,19 @@ def get_trees():
     return trees
 
 
+@bp.route('/compat/trees/')
```
Callek: in order to support /compat/trees itself I think we need to drop the trailing / (noteworthy the /trees above doesn't have trailing slash)

Callek:
I'm also partial to "v0" instead of "compat" when it comes to "maintain compat with old api version" needs.

But I feel only strongly enough to say so, and defer to your judgement on that one.

:+1: with the trailing slash issue recognized and addressed if needed.

relengapi/blueprints/treestatus/__init__.py
djmitche: The Apache proxy configuration always includes the trailing slash (since it's proxying from the root of https://treestatus.mozilla.org).

djmitche:
Fair enough, I'll use `v0`.  I'll also allow the missing trailing slash with another `@bp.route`.

djmitche:
how's that?

Callek:
:+1:

PR #327, opened 2015-08-28
Add tree status colors

djmitche:
These were likely lost accidentally while deleting a slew of unused CSS
rules.


Callek:
:sake: Sure :-)

PR #328, opened 2015-08-28
Make the tree log time a little easier to read

djmitche:
Not sure why I didn't notice this before, but the table cell was running
into its neighbors, making it difficult to read.  This adds a little
more breathing room, and moves the timezone (UTC) up to the header so
it's not repeated on every line.


Callek:
:+1:

PR #330, opened 2015-09-01
Set and verify MySQL connection charset

djmitche:
This will at least catch the case where the client charset is `latin1`.

(the PR will reduce coverage since we don't test against MySQL in Travis)

### DIFF
--- a/relengapi/lib/db.py
+++ b/relengapi/lib/db.py
@@ -126,6 +126,15 @@ class Alchemies(object):
         def foreign_keys_on(dbapi_con, con_record):
             # Set the default storage engine in case we create tables
             dbapi_con.cursor().execute("SET default_storage_engine=InnoDB")
+            # verify that the connection is utf-8
+
+        @event.listens_for(engine, "connect")
+        def check_client_charset(dbapi_con, con_record):
+            curs = dbapi_con.cursor()
+            curs.execute("show variables like 'character_set_client'")
+            
+            if curs.fetchone()[1] != 'utf8':
+                raise RuntimeError("MySQL client connection is not using utf8 encoding")
 
         return engine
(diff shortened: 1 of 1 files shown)


djmitche:
Oh, hey, we *are* using MySQL in Travis.  And not using utf8!

djmitche:
"We know that some of our users want MySQL 5.6 desperately, so select images (Ruby, PHP, Python, Node.js) for this 2015-04-27 updates will be available for those who wish to use it upon request. Please email us at support@travis-ci.com with the name of the repository you wish to use. (For this, you will have to use the standard infrastructure.)"

I have an idea (*set* the variables, thereby overriding the default)

djmitche:
Setting the variables doesn't help, since MySQLdb doesn't know about the changes.

Calling `dbapi_con.set_charset("utf8")` works, but it's fairly specific to MySQLdb.

This option, adding `charset=utf8` to the URL, works and seems fairly generic.

relengapi/lib/db.py
```diff
@@ -120,12 +120,25 @@ def foreign_keys_on(dbapi_con, con_record):
...
     def create_mysql_engine(self, url):
         url.query['use_unicode'] = "True"
+        url.query['charset'] = "utf8"
         engine = sa.create_engine(url)
 
         @event.listens_for(engine, "connect")
         def foreign_keys_on(dbapi_con, con_record):
             # Set the default storage engine in case we create tables
             dbapi_con.cursor().execute("SET default_storage_engine=InnoDB")
+            # verify that the connection is utf-8
```
Callek: drop this misplaced comment, please

Callek:
:+1: once you fix pep8

PR #326, opened 2015-09-01
Add a root page widget for treestatus

djmitche:


djmitche:
Ready for review (this was an easy one!)

Callek:
:+1: but I'd personally like to not have this live until we switch over to relengapi for treestatus, due to possible confusion

PR #323, opened 2015-09-02
Use real HTTP verbs in treestatus

Callek:
For example, `REVERT` is not a standard verb.


Callek:
ready for review @djmitche

Callek:
Note coveralls is failing "patch" because one of the lines that was caught in my "need to indent" was previously untested.

djmitche:
Looks good so far, but needs fixes to the frontend code to use the new method/argument.

Callek:
@djmitche this UI is untested, but aiui this is how it should be done.

djmitche:
yup, works for me

PR #332, opened 2015-09-02
Bug 1192021: log directly to mozdef, rather than via syslog

djmitche:
This switches things around substantially: non-mozdef output is now
not JSON encoded, so it will be easier to read, and only messages tagged
with `mozdef=True` will be sent to mozdef.  This is currently limited to
tooltool and tokenauth.


Callek:
:+1:

PR #334, opened 2015-09-03
Use UTCDateTime for treestatus

djmitche:
..and add some tests so nobody makes the same mistake again!

### DIFF
--- a/relengapi/tests/test_column_types.py
+++ b/relengapi/tests/test_column_types.py
@@ -1,0 +1,20 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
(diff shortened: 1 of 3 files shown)


relengapi/tests/test_column_types.py
```diff
@@ -0,0 +1,20 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import sqlalchemy as sa
+
+from nose.tools import eq_
+from relengapi.lib import db
+
+
+def test_no_DateTime_columns():
+    """No DB columns are of SQLAlchemy type `DateTime`"""
```
Callek: I'd love a "see issue<>" link here, so newcomers can know "why".

Callek:
:+1:

relengapi/tests/test_column_types.py
djmitche: It didn't really come down to a single issue, but I can include a little "why" in the assertion message.

PR #335, opened 2015-09-03
Set LOGIN_DISABLED explicitly to False in the test context

Callek:
This is to solve https://github.com/maxcountryman/flask-login/commit/2fc6a7e06b560233eadb44f8eba74e2a1bce88be on our local end, and to make sure that tests which care to verify that login is required will be able to pass accurately.

### DIFF
--- a/relengapi/lib/testing/context.py
+++ b/relengapi/lib/testing/context.py
@@ -47,6 +47,7 @@ class TestContext(object):
             return self._app
         config = self.options.get('config', {}).copy()
         config['TESTING'] = True
+        config['LOGIN_DISABLED'] = False  # Make @login_required enforced in tests.
         config['SECRET_KEY'] = 'test'
         config['SQLALCHEMY_DATABASE_URIS'] = uris = {}
         dbnames = self.options.get('databases', [])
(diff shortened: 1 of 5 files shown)


Callek:
ready for review.

relengapi/lib/testing/context.py
```diff
@@ -68,6 +69,8 @@ def _make_app(self):
             @app.before_request
             def set_user():
                 auth.login_manager.reload_user(user)
+        # Don't issue a 302 redirection when login is needed.
+        auth.login_manager.login_view = None
```
djmitche: I don't like this for two reasons
 * it's a permanent (not per-test) change
 * it changes the behavior of the API (as evidenced by changes to `test_lib_permissions.py`)

I can see two other options:
 * make this setting an argument to TestContext that can be enabled on a per-test or per-file basis (and revert it when the test is complete)
 * write a `check_login_required(resp)` test helper which verifies that resp is a 302 to the login page

Callek:
How's this?

djmitche:
Very nice!  Just needs an update to `relengapi/docs/development/testing.rst` and good to push

PR #273, opened 2015-09-09
Consider using REST path semantics better

Callek:
_From @Callek on October 23, 2014 23:31_

From Issue #4:

> * Regarding REST endpoints, I agree - we should try to use verbs where possible, and also to keep semantic meaning with the path. So I'd prefer GET /loans?inactive=1 to GET /loans/all, and creating a loan (whether for yourself or someone else) should be via POST /loans.

_Copied from original issue: mozilla/build-relengapi-slaveloan#6_


Callek:
@djmitche what say you?

relengapi/blueprints/slaveloan/rest.py
```diff
@@ -82,26 +82,11 @@ class UpdateManualAction(wsme.types.Base):
...
-    #: If known in advance, fqdn of the machine to loan
-    fqdn = unicode
-    #: If known in advance, ip address of the machine to loan
-    ipaddress = unicode
-
-
 class LoanRequest(wsme.types.Base):
     "Represents a new loan request"
 
+    #: Initial Status (Must have slaveloan.admin perm, without throws a Forbidden)
```
djmitche: I'm not sure what this means -- why mention a required permission on an attribute, rather than on the API call that requires it?

djmitche: Ah, I see, Must have .. to set this attribute.  Still, better indicated in the API call

djmitche:
Otherwise, looks good.  I skimmed the code enough to see that it combines the two loan request methods, and changes "get all loans".  I love the tests :)

PR #338, opened 2015-09-09
Bugfix - AutoType should allow newlines for type comments

Callek:
I noticed that docs were getting truncated, e.g.

https://api.pub.build.mozilla.org/docs/usage/tokenauth/#types
> **typ** (*Enum(prm, tmp, usr)*) -- token type (short string). This defaults to `prm` for backward

This patch makes that:
> **typ** (*Enum(prm, tmp, usr)*) -- token type (short string). This defaults to `prm` for backward compatibility, but should always be specified.

There are many other places in the code this is an issue


Callek:
@djmitche I had no idea how to test this, but if you have any easy ways I'm open.  Either way, good-to-merge?

djmitche:
No need to test -- if the result looks good, it looks good :)

Issue #276, opened 2015-06-05
fiximports.py shouldn't rewrite the file when it doesn't change it


djmitche:
_From @Callek on October 14, 2014 21:48_

In testing I have the case where I get "file was modified..." warnings when I have an editor open and I run validate.sh.

We should adjust fiximports.py to not do an open(.., 'w') unless it needs to.  CC: @tardyp who may be interested in backporting this change elsewhere.

_Copied from original issue: mozilla/build-relengapi-skeleton#2_

djmitche:
Maybe another option is to modify fiximports.py to have a "detect" mode and a "fix" mode.  Then we can write ./fixup.sh to run fiximports.py in fix mode *and* run autopep8 with appropriate options.

djmitche:
_From @mrrrgn on October 15, 2014 15:34_

Why not just add a dry-run option?

djmitche:
Same idea, yeah.  I just want to avoid manually futzing with imports (and whitespace for pep8) when the check fails.

djmitche:
_From @Callek on October 15, 2014 21:17_

fwiw I just found https://github.com/miki725/importanize which even itself has a lot of changes to relengapi base. I didn't verify yet if its actually better than fiximports in adhering to pep8.

PR #302, opened 2015-09-09
Suggest and require absolute imports

Callek:
We use absolute imports just about everywhere.  They're easier to read, as you're not left searching through a list of directories for the relevant file.  We should include that in the patch requirements doc and (not sure how) enforce that in validate.sh.


Callek:
I solved this by using isort, which allows me to easily add the import to every py file, in the proper way.

Along with doing this, we get better "import enforcement" (with some actual changes to go along with it) in a way that doesn't forcibly touch files when you run it.... It also allows us to output a diff on checking, and enforce the `from __future__ import absolute_import` here.

@djmitche r?  (the best way to review this is probably sha by sha, rather than as one big hunk)

Callek:
Also fixes #276

djmitche:
I like it!

djmitche:
One thought: it might be nice if the comment included a way to fix the ordering when validate fails -- that will help contributors who get a red X from travis.

PR #339, opened 2015-09-10
Only isort relengapi/

djmitche:
@Callek - the `-rc .` was picking up my virtualenv, and that has lots of unsorted files in it :)

### DIFF
--- a/validate.sh
+++ b/validate.sh
@@ -88,12 +88,12 @@ start_step "running pyflakes"
 pyflakes relengapi || not_ok "pyflakes failed"
 finish_step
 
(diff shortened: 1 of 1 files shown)


Callek:
I'd love if we also isort'd the setup.py but thats less of an issue to me (we shouldn't really be adding much of any imports there at all)

Admittedly I treat my path as <venv_root>/{bin,repo_root,lib,...} rather than venv_root==repo_root  so I didn't hit that issue. -- also huh at travis failing on my slaveloan tests.....

:+1:

djmitche:
Yeah, dunno what's up with those slaveloan tests..

PR #304, opened 2015-09-10
Upgrade AngularJS

djmitche:
The version of AngularJS we're using has a bug where it doesn't unCamelCase attribute names correctly when specified in isolate scopes.


djmitche:
Probably not much to "review" here, but worth downloading and playing with to make sure I didn't miss any major functionality breakage..

Callek:
Reading the changelog, possible issue:

>ngRepeat
>
>Due to c260e738, previously, the order of items when using ngRepeat to iterate over object properties was guaranteed to be consistent by sorting the keys into alphabetic order.
>
>Now, the order of the items is browser dependent based on the order returned from iterating over the object using the for key in obj syntax.
>
>It seems that browsers generally follow the strategy of providing keys in the order in which they were defined, although there are exceptions when keys are deleted and reinstated. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Cross-browser_issues
>
>The best approach is to convert Objects into Arrays by a filter such as https://github.com/petebacondarwin/angular-toArrayFilter or some other mechanism, and then sort them manually in the order you need.

Callek:
Docs, ftr: https://docs.angularjs.org/guide/migration

and I grant :+1: if you are satisfied that we don't want to care about the ng-repeats in the code that don't also have orderBy set.

djmitche:
I don't think we ever iterate over objects, so that should be OK.

PR #344, opened 2015-09-12
docs: Some documentation missing '/mapper' prefix

hwine:
The example URLs were all correct, but the headings were not.


Callek:
:+1: merging now

PR #345, opened 2015-09-13
mock-1.3.0 install is buggy

djmitche:
If you run against the lastest mock, it fails because funcsigs isn't installed - see https://github.com/testing-cabal/mock/issues/316.

Easiest fix is to temporarily require funcsigs ourselves.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
     extras_require={
         'test': [
             'nose',
-            'mock',
+            'funcsigs',  # https://github.com/mozilla/build-relengapi/issues/345
             'coverage',
             'pep8',
             'mockldap',


Callek:
I'm on my mobile right now, and don't have my github pw handy. But +1 for
this.

rbtcollins:
I suspect you have an old pip actually.

djmitche:
@rbtcollins -- why's that? Do newer pips read requirements.txt?

djmitche:
Indeed, you're right.  Not sure how that works!

I'll back this out, then.

rbtcollins:
The requirements.txt file contents is not read by pip - its reflected into setup kwargs by pbr.

What happens here is that pip 6 introduced PEP 426 marker support. The funcsigs conditional dependency is expressed via markers, and thus pip below 6 will appear to not know about the dependency. Sadly there isn't anything I can do [yet] to make this fail early - on the one hand wheels supported markers from the first release, via the setuptools syntax for them, but because pip < 6 can install wheels, the wheel of mock has its marker based dependencies ignored. If I remove the wheel, then setup.py could in principle look for what pip is running, but I haven't found a reliable way to detect that. Its something we're going to flesh out in the revised marker PEP underway at the moment.

PR #301, opened 2015-09-16
restapi Angular service should set Content-Type on requests

djmitche:
I just had to add `headers: {'Content-Type': 'application/json'}` to a restapi request; restapi should do this for me when there is data present.

### DIFF
--- a/relengapi/static/js/relengapi.js
+++ b/relengapi/static/js/relengapi.js
@@ -9,6 +9,11 @@ angular.module('relengapi').config(function($httpProvider) {
         var url = config.url;
         return meth + " " + url;
     };
+
+    // Unlike POST and PUT, Angular doesn't set the content-type for 'patch' by
+    // default, but we'd like it to do so
+    $httpProvider.defaults.headers.patch['Content-Type'] = $httpProvider.defaults.headers.post['Content-Type'];
+
     $httpProvider.interceptors.push(function($q) {
         return {
             'request': function(config) {
@@ -16,7 +21,7 @@ angular.module('relengapi').config(function($httpProvider) {
                     if (config.data) {
                         // Firefox will helpfully produce a clickable rendition of the data
                         console.log("RelengAPI request:", summarize_config(config),
-                                    'body', JSON.parse(config.data));
+                                    'body', config.data);
                     } else {
                         console.log("RelengAPI request:", summarize_config(config));
                     }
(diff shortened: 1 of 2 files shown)


djmitche:
`$http.patch` does exist in the newer version of Angular.  So this is ready for review.

relengapi/static/js/relengapi.js
```diff
@@ -82,20 +82,24 @@ angular.module('relengapi').provider('restapi', function() {
             /* add the flag that our interceptor uses to identify RelengAPI requests */
             config.is_restapi_request = true;
 
+            /* add content-type */
+            config.headers = config.headers || {};
+            config.headers['content-type'] = 'application/json';
```
Callek: This will blow away other explicitly defined content-type headers... which luckily we don't have any examples of, but it could cause some problems with charsets!

https://github.com/mozilla/build-relengapi/blob/b6079a7e9421a9bd79b6113802efed83a8b186c3/relengapi/blueprints/tokenauth/static/tokens.js#L117

Perhaps setting charset here would be fine, and/or a check on if content-type is already defined?

relengapi/static/js/relengapi.js
```diff
@@ -65,7 +65,7 @@ angular.module('relengapi').config(function($httpProvider) {
 });
 
 angular.module('relengapi').provider('restapi', function() {
-    var wrap = function(wrapped) {
+    var wrap = function(wrapped, has_data) {
```
Callek: has_data is unused.

djmitche:
I'm not sure where the charset=utf8 thing came from, TBH.  Angular adds it in its defaults
```
var CONTENT_TYPE_APPLICATION_JSON = {'Content-Type': APPLICATION_JSON + ';charset=utf-8'};
...
    // default headers
    headers: {
      common: {
        'Accept': 'application/json, text/plain, */*'
      },
      post:   shallowCopy(CONTENT_TYPE_APPLICATION_JSON),
      put:    shallowCopy(CONTENT_TYPE_APPLICATION_JSON),
      patch:  shallowCopy(CONTENT_TYPE_APPLICATION_JSON)
    },
```
so that should already be working, and probably is.  I think the restapi call I was making when I filed this bug was
```
        restapi({
            url: '/tokenauth/tokens',
            method: 'POST',
            headers: {'Content-Type': 'application/json; charset=utf-8'},
```
which doesn't get the default headers for `POST` because it's not using `restapi.post` (which calls `$http.post`).  Also, the defaults aren't set for `PATCH`.

So, I think fixing this for `PATCH` and calling `restapi.post` in `token.js` are really the only two things to do in this bug.

djmitche:
Let's see how that does..

Callek:
assuming its been tested, :+1:

Issue #157, opened 2014-12-09
Docs have no indexes


djmitche:
This is actually a [Sphinx bug](https://bitbucket.org/birkenfeld/sphinx/issue/1003/using-websupport-how-do-i-get-the), so we wait until that's fixed.

hwine:
sphinx move to github, blocking issue is now sphinx-doc/sphinx#1003

Callek:
When the sphinx bug is fixed, we should backout PR #350

PR #350, opened 2015-09-22
Mask issue #157 by not displaying index or module links

Callek:
Stop confusing people by nature of having an empty index.


djmitche:
lgtm

Issue #351, opened 2015-10-07
Bugzilla actions should use apikey instead of raw password


Callek:
So, our code currently uses user+password for bugzilla authorization.

https://github.com/mozilla/build-relengapi/blob/master/relengapi/blueprints/slaveloan/bugzilla.py

This apikey support was added into https://github.com/bhearsum/bzrest itself, and is available in bzrest 1.1, so we will need to bump the requirements:
https://github.com/mozilla/build-relengapi/blob/master/setup.py#L40

We also need to modify docs https://api.pub.build.mozilla.org/docs/deployment/slaveloan/#bugzilla

Once this piece deploys we need to modify the deployment config, and make sure it has the apikey (which is already generated and in our private repo)

ciobancai:
Created pull request for the above changes #352

Callek:
fixed by #352

PR #352, opened 2015-10-13
First bug  #351

ciobancai:


relengapi/docs/deployment/slaveloan.rst
```diff
@@ -6,11 +6,9 @@ Bugzilla
 
 Slaveloan makes extensive use of Bugzilla's REST API for classifying new loans, reopening slave tracking bugs, etc.
 
-In order to use it, you *must* specify where and what user to use (in the settings file). e.g.::
+::
 
     BUGZILLA_URL = "https://bugzilla-dev.allizom.org/rest/"
-    BUGZILLA_USER = "my.user@example.com"
-    BUGZILLA_PASS = "my.passw0rd"
```
djmitche: This doesn't seem to say anything about the API key, and also removes docs about BUGZILLA_USER and BUGZILLA_PASS (which, I think, will still work, right?)

djmitche:
It'd be best to rebase and remove 87f4be3d62f7c84aaf7a96e1935c1094dd63460f and 23cca6b since they are just test commits.

Callek:
At this point I'd also rebase to remove the docs + revert-of-docs ;-)

That said, we do need to update the docs, but they should reflect the code changes you do.

The function that needs changing is https://github.com/mozilla/build-relengapi/blob/master/relengapi/blueprints/slaveloan/bugzilla.py#L21 which explicitly reads the set config values and configures bzrest to use those values. Without that change this does nothing.

setup.py
```diff
@@ -23,7 +23,7 @@
         "Celery>=3.1.16",  # see https://github.com/mozilla/build-relengapi/issues/145
         "alembic>=0.7.0",
         "argparse",
-        "requests",
+        "requests>=1.2.3",
```
Callek: I am similarly unsure why you're bumping requests version here.

relengapi/blueprints/slaveloan/bugzilla.py
```diff
@@ -17,17 +17,13 @@
 MAX_ALIAS = 15
 DEFAULT_BUGZILLA_URL = "https://bugzilla-dev.allizom.org/rest/"
 
-
```
Callek: relengapi/blueprints/slaveloan/bugzilla.py:20:1: E302 expected 2 blank lines, found 1

Callek:
Also needs a doc update, to describe the new config var.

relengapi/docs/deployment/slaveloan.rst
```diff
@@ -9,8 +9,7 @@ Slaveloan makes extensive use of Bugzilla's REST API for classifying new loans,
 In order to use it, you *must* specify where and what user to use (in the settings file). e.g.::
```
Callek: s/and what user to use/and an apikey for a user/

setup.py
```diff
@@ -37,8 +37,7 @@
         "IPy",
         "furl",
         "redo",
-        # Temporary freeze until https://github.com/bhearsum/bzrest/pull/3 is fixed
-        "bzrest==0.9",
+        "bzrest==1.1",
```
Callek: I'd actually like >=1.1 here, incase there are other updates that are useful.

Callek:
:+1: \o/ Thanks for the patch

PR #358, opened 2016-01-19
Bug 1204758: stream mapfiles rather than building them in memory

djmitche:
@hwine do you want to have a look?

### DIFF
--- a/relengapi/blueprints/mapper/__init__.py
+++ b/relengapi/blueprints/mapper/__init__.py
@@ -106,20 +106,24 @@ def _project_filter(projects_arg):
         return Project.name == projects_arg
 
 
-def _build_mapfile(query):
+def _stream_mapfile(query):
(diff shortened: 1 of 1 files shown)


djmitche:
Note that the tests only verify that we're getting the right result -- I don't know of a way to convince the MySQL libraries to indicate when they are actually downloading the data, but I assume that they stream when possible.

hwine:
> Note that the tests only verify that we're getting the right result -- I don't know of a way to convince the MySQL libraries to indicate when they are actually downloading the data, but I assume that they stream when possible.

A quick way to verify is to curl the full mapfile for gecko-dev. Currently, results do not start to get returned until about 15-20 seconds in.

djmitche:
Let's do that once this lands.  I don't think the staging DB has a full mapfile, so we'll need to wait for prod.

hwine:
r+ -- I'll let you merge as you deploy.

PR #361, opened 2016-02-02
Bug 1245170: Allow tooltool uploads with non-user-associated tokens

djmitche:
Still needs tests; see https://bugzilla.mozilla.org/show_bug.cgi?id=1245170 for the motivation.

### DIFF
--- a/relengapi/blueprints/tooltool/__init__.py
+++ b/relengapi/blueprints/tooltool/__init__.py
@@ -135,7 +135,9 @@ def upload_batch(region=None, body=None):
     try:
         body.author = current_user.authenticated_email
     except AttributeError:
-        raise BadRequest("Could not determine authenticated username")
+        # no authenticated_email -> use the stringified user (probably a token
+        # ID)
+        body.author = str(current_user)
 
     # verify permissions based on visibilities
     visibilities = set(f.visibility for f in body.files.itervalues())
--- a/relengapi/blueprints/tooltool/test_tooltool.py
+++ b/relengapi/blueprints/tooltool/test_tooltool.py
@@ -56,6 +56,17 @@ TWO_DIGEST = hashlib.sha512(TWO).hexdigest()
 NOW = 1425592922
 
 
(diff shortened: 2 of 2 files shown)


relengapi/docs/relnotes/index.rst
```diff
@@ -2,6 +2,7 @@ Release Notes
 =============
 
 .. toctree::
+    3.3.4
```
Callek: Looks like you never merged 3.3.4 into master, I don't see why it should be part of this PR...

Callek:
I'm also confused on the relative spurious use of newlines in this PR, its as if select lines were autopep8'd without adjusting the --max-line-length. I'd personally rather we avoid the [seemingly] spurious changes.

relengapi/docs/relnotes/index.rst
djmitche: Oops -- pushed to master.  @garbas this is why we should have CD :)

djmitche:
Sorry about the line wraps -- I didn't mean to!

Callek:
offhand I'm not sure if the test supports the Token auth type (as in, if we use a token we're not necessarily testing that tokenauth works here -- just a 'special' user auth, so that might be a room for testing improvement given the goal of this PR)

That said, I don't think blocking on it is worthwhile at all, so :shipit:

garbas:
@djmitche code looks straightforward +1 on that, but I'm not sure I understand the whole picture to be able to judge how this affects other things.

relengapi/docs/relnotes/index.rst
garbas: Quoting Dustin J. Mitchell (2016-02-03 16:11:41)
> In relengapi/docs/relnotes/index.rst:
> 
> > @@ -2,6 +2,7 @@ Release Notes
> >  =============
> >
> >  .. toctree::
> > +    3.3.4
> 
> Oops -- pushed to master. @garbas this is why we should have CD :)
> 

give me a week and i'll get it done! :)

PR #362, opened 2016-02-15
Bug 1248205 - Don't allow changing the motd for more than one tree at a time

KWierso:
Not really tested much, but I think this should work. 

The change to treeStatusControl.html should hide the MotD elements on the main treestatus page unless only a single tree is selected selected (it won't show without first selecting a single tree), while still leaving it visible on the individual tree pages.

The change to treestatus.js should prevent mass clobbering if you select a single tree on the main page, type a MotD into the form, then continue to select additional trees before clicking "Update".

I was going to use an ng-if instead of ng-hide, but that seemed to break updating MotDs entirely.

### DIFF
--- a/relengapi/blueprints/treestatus/static/treestatus.js
+++ b/relengapi/blueprints/treestatus/static/treestatus.js
@@ -147,8 +147,8 @@ function(allowed_statuses, allowed_tags, restapi) {
                 var update = {trees: scope.trees};
 
                 // only update the MOTD if it's nonempty (which precludes empty
-                // MOTD's, but that's OK)
-                if (scope.message_of_the_day) {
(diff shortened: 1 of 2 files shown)


relengapi/blueprints/treestatus/static/treestatus.js
```diff
@@ -148,7 +148,7 @@ function(allowed_statuses, allowed_tags, restapi) {
 
                 // only update the MOTD if it's nonempty (which precludes empty
                 // MOTD's, but that's OK)
```
djmitche: mind extending this comment a little?

djmitche:
This looks good -- if you update that comment, I'll merge and ship.

KWierso:
Squashed in a comment update.

PR #365, opened 2016-02-25
clobber: added Taskcluster purge cache simple form

garbas:


relengapi/blueprints/clobberer/__init__.py
```diff
@@ -255,6 +257,50 @@ def forceclobber():
     builddir = request.args.get('builddir')
     return "{}:{}:forceclobber".format(builddir, future_time)
 
+
+def get_tc_credentials(config):
+    credentials = []
```
djmitche: Why a list?

djmitche:
This works, although as written it needs tests 

Ideally, though, we're not making TC requests on behalf of the user (although it's nice ot see that these are authenticated with relengapi permissions).  Instead, we'd be making those calls directly from the JS frontend.  Fixing https://bugzilla.mozilla.org/show_bug.cgi?id=1246191 should have helped a little, as you don't need credentials to reveal the list of workerTypes.

https://bugzilla.mozilla.org/show_bug.cgi?id=1250160 will fix the errors in Travis.

garbas:
@djmitche can i ask for another review of this PR

relengapi/blueprints/clobberer/rest.py
```diff
@@ -37,3 +37,31 @@ class ClobberTime(wsme.types.Base):
...
+class TCWorkerType(wsme.types.Base):
+    """Represents worker type of taskcluster
+    """
+
+    name = wsme.types.wsattr(unicode, mandatory=True)
+    caches = wsme.types.wsattr([str], mandatory=False, default=[])
+
+
+class TCBranches(wsme.types.Base):
+    """Represents branches of taskcluster
```
djmitche: An instance of this represents a single branch, though -- so perhaps TCBranch is a more appropriate name?

relengapi/blueprints/clobberer/static/clobberer.js
```diff
@@ -81,4 +83,80 @@ angular.module('clobberer').controller('ClobberController',
...
+      return workerTypes;
+    };
+
+    $scope.loadBranches = function() {
+        $scope.TCBranches = null;
+        $scope.selectedTCBranch = null;
+        $scope.selectedTCWorkerTypes = {};
+
+        restapi.get('/clobberer/tc/branches').then(function (data) {
+            $scope.TCBranches = data.data.result;
```
djmitche: This should be wrapped in `$scope.$apply(function() { .. })`

djmitche:
If we "nixify" relengapi, let's do it in a separate PR.  Not that I'm opposed, I just need to learn it, and I don't want to confuse two unrelated things.

I'm still not convinced that clobbering all caches is the right decision, but ultimately it's just wasted Mozilla engineering time and money in Amazon's pocket, and I don't have metrics for how much, so I'll stop arguing.

The "Buildbot builders" section currently shows "loading" until you choose a branch, which was not the case for the previous version.  I thin that should only appear *after* a branch is selected from the <select>, and before that branch's last-clobber information is loaded.

It would be nice to get 100% of the python code in the diff covered -- I see that the exceptions are use of TC credentials and handling an exception.  I don't think this should be too hard to accomplish with the patched `taskcluster.PurgeCache.purgeCache` that you are using in the tests.

Finally, on the issue of system security: there's a small "[confused deputy](https://en.wikipedia.org/wiki/Confused_deputy_problem)" problem here: RelengAPI only knows of the "clobberer.post.clobber" permission, but is using the full access accorded to the given TC credentials to satisfy that request.  Many users should really only have permission to clobber level-1 (try) or level-2 (twigs) repositories, but not level-3 (mozilla-central, releases), but we don't have the resolution here to make that determination.  Even if we did, the code translating relengapi permissions into TC scopes would need to be examined carefully as it would be a security sensitive "deputy".

If RelengAPI, instead of using LDAP auth and "permissions", accepted TaskCluster credentials and handled scopes, then we would not have to implement any kind of deputy, nor would RelengAPI need its own credentials.  Instead, the frontend code would make the purgeCache API requests directly to the PurgeCache service, using the user's own credentials.

Ultimately, this security issue is not a big deal and there's no need to try to fix it now: clobbering can only slow things down, so there's no risk of malicious content injection into build processes or anything like that.

garbas:
@djmitche could you give another view? i renamed TCBranches -> TCBranch, use $scope.$apply and removed nix related stuff. i'll submit that in a separate PR after this one gets merged.

djmitche:
Ugh, I was wrong about the `$scope.$apply` bit -- the $http service relies on $q which takes care of that.

djmitche:
Looks like `VERSION` got omitted from the gardening commit?

The purging works well.  I missed the `TASKCLUSTER_CACHES_TO_SKIP` config -- it should be documented in a new `relengapi/docs/deployment/clobberer.rst`, along with the new permissions.

I suspect the buildbot "loading" issue is actually because I have zero branches in my DB.  So nothing to fix there.

In the interest of expediency, I'll get this fixed up and merged and deployed, including docs.

djmitche:
(since I believe you're at orientation today?)

garbas:
@djmitche tnx fixing and merging. is there anything left to do here?

djmitche:
Nope, I think that's it, and Tomcat has given a :thumbsup:.  Thanks!

PR #366, opened 2016-03-02
Bug 1249414: Handle error other than BadData in processing signature

djmitche:


djmitche:
It turns out this isn't an issue with itsdangerous 0.24, but it's worth keeping the `except` and the test around.

garbas:
:+1: 

@djmitche does it makes sense to add test for that new exception?

djmitche:
I tried to, but as the coverage shows, with the upgraded version of `itsdangerous`, that branch is not reached.  Still, I think it's a good thing to have in place.  I'll mark the `except` block as not being covered, and also change the comment.

djmitche:
updated - have another look?

djmitche:
The travis errors appear unrelated to this particular change.

garbas:
:100:

PR #369, opened 2016-03-22
nixify relengapi

garbas:
Purpose of this PR is to introduce Nix as a build a tool to help setup development platform. This does not remove current developer workflow using ``pip/virtualenv``, but only extends it.

And why would somebody consider using Nix?

## Atomic operations

This is a fancy word that makes sure you never end up with a broken environment. ``pip install --upgrade`` (or ``npm install`` or ``apt-get upgrade``) is known to sometimes break your environment which causes developers to lose time on maintaining their environment rather then solving an assigned problem. 

Nix commands are atomic by its design. Only once everything was successfully downloaded and installed, Nix switches to new environment. If something fails during installation 

## Works alongside your system package manager

Usually adoption of new technology comes in "All or nothing" form. Meaning you need to use it all the way or not at all. This is especially true for package managers.

Nix is different. Nix won't write to ``/usr`` and because of this it can work alongside any other package manager. Nix installs everything inside ``/nix/store`` folder and in worse case, when you are unhappy with Nix, you can simple remove Nix (and every package installed by Nix) from your system by deleting ``/nix`` folder. Until you are ready to give it another try :smile: 

## System dependencies

``pip`` is a language specific package manager and handles only packages at language specific level. Even a simple application such as ``relengapi`` have some system dependencies (eg. python-ldap, ...).

Nix can install all dependencies, including system one. Your project can directly depend on specific python version, nodejs for frontend work, openldap, ... you name it.

## Complete dependency tree

In a group where developers share development environment we quite often hear "It works on my laptop!". Reproducability of environments between systems is a hard problem and Nix excels at this.

Nix forces you to specify **ALL** of your dependencies, including system one. While this is a bit cumberstone to do, it is the **only** way to make sure others can reproduce environment on their machines. Also only one person has to do that initially and then everybody benefits.

## Cross platform / Binaries

Nix works across platforms Linux 32/64, FreeBSD, Darwin/OSX, Windows/Cygwin. This PR was only tested on Linux 64.

Nix ecosystem comes with CI server which you can setup yourself (like I did, http://hydra.garbas.si). This way you can create binaries (for custom packages) which can be also used both in production and during development. Since binaries are reused between production and development, you can catch many installation problems before deploying to production.

## Nix is easy to use

A typical developer does not need to know anything about Nix to be able to used it. Once Nix is installed one need to only know 2 commands:

 - ``nix-shell`` to enter development environment
 - ``nix-build`` to perform build of the project (the same command is used in CI and later in production)

When changes in Nix setup are needed, a  minimum knowledge of Nix is needed. Also Nix is a DSL sized language which can be learned in less then a day.

## Nix is not perfect

I don't want to fool you that Nix is perfect. It definitely can be improved and many improvements are in the making. 

 - While there is a very good manual for every piece of code newcomer documentation (eg. Getting started chapter) is clearly missing.

 - Command line utilities *feel* a bit old. Revamp of command line UX is in making.

 - While it might feel you are learning yet another tool (which is also true) you will soon realize that you will stop using many other package managers (no more pip/npm/apt-get/....).

# Give it a try

1. Install Nix using bellow command

  ```sh
  % curl https://nixos.org/nix/install | sh
  ```

2. Clone ``feature/nixify`` branch

  ```sh
  % git clone https://github.com/garbas/mozilla-relengapi
  % cd mozilla-relengapi 
  mozilla-relengapi/ % git checkout -b feature/nixify origin/feature/nixify
  ```

3. Enter development environment and run tests

  ```sh
  mozilla-relengapi/ % nix-shell
  (nix-shell) mozilla-relengapi/ % relengapi run-tests
  ```

  press CTRL-D to exit.

4. Build relengapi script

  ```sh
  mozilla-relengapi/ % nix-build
  mozilla-relengapi/ % ./result/bin/relengapi --help
  ```

 ## Ask questions in comments below


garbas:
@djmitche hope this explains Nix features a bit. who you think i should share with to get feedback (dont want to spam unnecessary)

.travis.yml
```diff
@@ -11,10 +11,7 @@ env:
 # command to install dependencies
 install:
   - "pip install -e .[test]"
-  - "pip install coverage"
-  - "pip install codecov"
   - "pip freeze"
-  - "pip install MySQL-python"
```
djmitche: Why remove these from travis?  It doesn't look like Travis will be using nix..

default.nix
```diff
@@ -0,0 +1,246 @@
...
+        propagatedBuildInputs = with self; [ ipython ];
+      }) //
+    (buildPythonPackage "ipython" "4.0.3" "16d4c8e79510ba427fb5336e15b0ea34" {
+        LC_ALL="en_US.UTF-8";
+        buildInputs = with self; [ nose pkgs.glibcLocales Pygments mock ];
+        propagatedBuildInputs =
+          (with self; [ decorator pickleshare simplegeneric traitlets requests pexpect ]) ++
+          (with pythonPackages.python.modules; [ readline sqlite3 ]);
+        patchPhase = ''
+          sed -i -e "s|'\:sys_platform \=\= \"darwin\" and platform_python_implementation \=\= \"CPython\"': \['gnureadline'\],||" setup.py
```
djmitche: Why would relengapi need to know about this obscure patch to build ipython?

default.nix
```diff
@@ -0,0 +1,246 @@
...
+        propagatedBuildInputs = [ self."path.py" ];
+      }) //
+    (buildPythonPackage "pyflakes" "1.1.0" "e0bf854cd5abd4527e149692925b82eb" { }) //
+    (buildPythonPackage "python-dateutil" "2.5.1" "2769f13c596427558136b34977a95269" {
+        propagatedBuildInputs = with self; [ six ];
+      }) //
+    (buildPythonPackage "python-editor" "0.5" "ece4f1848d93286d58df88e3fcb37704" {}) //
+    (buildPythonPackage "python-ldap" "2.4.25" "21523bf21dbe566e0259030f66f7a487" {
+        NIX_CFLAGS_COMPILE = "-I${pkgs.cyrus_sasl}/include/sasl";
+        propagatedBuildInputs = [ pkgs.openldap pkgs.cyrus_sasl pkgs.openssl ];
```
djmitche: why no `with self`?

djmitche:
So a few concerns here, having looked through this a bit:

 * This seems to duplicate the information in setup.py/requirements.txt yet again into a nixfile.  In fact, it seems to reflect the entire dependency hierarchy of the Python package (which, happily, is much shorter than it would be with Node.js packages!), and to have some "hacks" to get some of those packages to build.  In other words, it seems like quite an additional maintenance burden to keep up with as the landscape of our dependencies changes, or as we add new dependencies.  At the least, we should have some automated testing to make sure all installation methods are in sync and functional.

 * Since this still relies on pypi, it doesn't seem to address some of the more problematic issues with that service:
   * if packages are deleted from pypi, this will fail
   * I don't see any way to identify available updates to packages, which is equally important to holding package versions fixed, especially for security-sensitive libraries like itsdangerous, Flask, and .. well, most of the things we're using here
   * These builds aren't byte-for-byte reproducible, are they?  The inputs are the same, but are compiler and library versions pinned?  Builds modified to not include timestamps?  Is there sandboxing to prevent rogue build processes from pulling in inputs from the Internet?
   * On the flip side, say another vuln appears in glibc.  We have tools to scan `/lib/libc*` and look for vulnerable versions, but would those miss nix-built binaries?  In other words, would we need to perform redeploys of all nix-built apps in addition to the `yum upgrade` or whatever that the folks maintaining the hosts do?

 * It's minor, but this uses a lot of space and is pretty slow -- my relengapi virtualenv is 155M, whereas /nix on my docker container is 1.3G.  It took something like 10 minutes to set up the nix environment (the first `nix-shell` run).

default.nix
garbas: ``with self;`` adds whatever is in self attr set (dict) to current scope. consider this two examples to be the same:

```nix
a_list = with self; [ A B C ];
```

```nix
a_list = [ self.A self.B self.C ];
```

.travis.yml
garbas: this is part of first commit (non nix related), but i included packages needed during testing at `[test]` extra in `setup.py`

default.nix
garbas: `ipython` is a dependency of `ipdb` which is only added as a helper tool during development. if you run `nix-build/nix-shell--arg develop false` if won't included in environment.

garbas:
> This seems to duplicate the information in setup.py/requirements.txt yet again into a nixfile. In fact, it seems to reflect the entire dependency hierarchy of the Python package (which, happily, is much shorter than it would be with Node.js packages!), and to have some "hacks" to get some of those packages to build. In other words, it seems like quite an additional maintenance burden to keep up with as the landscape of our dependencies changes, or as we add new dependencies. At the least, we should have some automated testing to make sure all installation methods are in sync and functional.
>

only partially. i already reuse requirements*.txt files to add files
as dependencies. which would break nix when somebody adds package into
requirements*.txt package. this is intended behavior since you don't
want to ignore this silently and everybody that uses nix will have to
fix it before continuing the work ensuring proper packages are listed.

also the list of those packages will be generated in the future using
pypi2nix tool (https://github.com/garbas/pypi2nix). i need to test
pypi2nix tool before i start promoting it and using it in projects.

those "hacks" are usually problems which system packages also need to
work around, since many of those hacks were actually ported from their
repositories. those hacks are just invisible to users, in nix we show
you exact commands which are used to build them since build system is
more transparent and you can trace all build steps for any package.
many of those hacks were already reported upstream, if relevant for
upstream (to source of a package).

garbas:
> Since this still relies on pypi, it doesn't seem to address some of the more problematic issues with that service:
>
> if packages are deleted from pypi, this will fail

source packages are kept on hydra (CI) or can also be backed up
manually (https://nixos.org/wiki/Download_all_sources).

> I don't see any way to identify available updates to packages, which is equally important to holding package versions fixed, especially for security-sensitive libraries like itsdangerous, Flask, and .. well, most of the things we're using here

if you will update requirements.txt nix-shell/nix-build will fail with
message explaining which package needs to be updated in nix. if
requirements.txt would provide all information to ensure deterministic
build this manual steps wouldn't be needed.

> These builds aren't byte-for-byte reproducible, are they? The inputs are the same, but are compiler and library versions pinned? Builds modified to not include timestamps? Is there sandboxing to prevent rogue build processes from pulling in inputs from the Internet?

no binaries are "only" build deterministic. this is a pre-requirement
before you can tackle byte-for-byte determinism. around ~40% of
packages are already byte-by-byte deterministic in nix, but there is
no man power to lead this effort further in nixos community. currently
debian effort is being followed and once debian is byte-by-byte
deterministic, nixos is the next day.

> On the flip side, say another vuln appears in glibc. We have tools to scan /lib/libc* and look for vulnerable versions, but would those miss nix-built binaries? In other words, would we need to perform redeploys of all nix-built apps in addition to the yum upgrade or whatever that the folks maintaining the hosts do?
>

yes, this is already done in nix (we just need to bump to latest, i
already do this for my machines) and redeploy the app. this can be
easily automated using hydra CI.

garbas:
> It's minor, but this uses a lot of space and is pretty slow -- my relengapi
> virtualenv is 155M, whereas /nix on my docker container is 1.3G. It took
> something like 10 minutes to set up the nix environment (the first nix-shell
> run).
>

- currently you have to build custom python packages manually. if you
would add http://hydra,garbas.si it would use binary packages from
there (i build them from develop branch of my fork)

- full closure size (all dependencies) of relengapi is currently 208M

   % nix-build
   % NIX_PAGER= du -scl $(nix-store -qR ./result) | grep "total$"
   208M    total

- to build docker image with only relengapi (and busybox) based on
scratch you can run

  % nix-build release.nix -A docker.x86_64-linux
  % docker load -i result/docker.tar.gz
  %  docker images
   REPOSITORY               TAG                 IMAGE ID
CREATED             SIZE
   docker-relengapi-3.3.7   3.3.7               03141a0bff52        46
years ago        334.9 MB
   % docker run docker-relengapi-3.3.7:3.3.7 relengapi --help

                                                            usage:
relengapi [-h] [--quiet] [--disable-logging]
    ...

or you can download docker image from http://hydra.garbas.si/build/132
(look for Build products)

- the full closure size it is going to be reduced even further once
this gets merged https://github.com/NixOS/nixpkgs/pull/7701

djmitche:
My take-away after reading these comment a few times is, "this doesn't make things any worse, and doesn't interfere with things non-nix users want to do".

 * Partially duplicates information in requirements.txt, with the addition of lots of package-specific hacks (which I maintain belong in descriptions of those packages, not in relengapi) and the full dependency tree (again, relengapi should not have to model this).  But this stuff is only of concern to Nix hackers, so I can blithely change `requirements.txt` and next time you run `nix-shell` you'll make a PR to update the `*.nix` files.
 * Builds verify hashes (well, md5's.. so, not worse than not validating), but aren't deterministic, so a little better than installing with pip from requirements.txt (and slightly worse than using peep)
 * Non-Python dependencies (ldap, mysql, etc.) are completely specified, so that's cool -- but I still feel like this is a solution in search of a problem.
 * We are on our own to handle any security upgrades in libraries we depend on, which are hidden from typical scanning tools.  So shipping these Non-Python dependencies increases our burden vs. just `apt-get install libmysql`.

So to use the Apache voting notation, I'm +0 on this -- I don't really see why this is useful, but you do, and I don't have any specific reason not to do it.

garbas:
@djmitche

just few clarifications for certain points, since i see i suck at
explaining :) but at least i'm getting slowly better. should i go
ahead and merge this?


> Partially duplicates information in requirements.txt, with the addition of lots of package-specific hacks (which I maintain belong in descriptions of those packages, not in relengapi) and the full dependency tree (again, relengapi should not have to model this). But this stuff is only of concern to Nix hackers, so I can blithely change requirements.txt and next time you run nix-shell you'll make a PR to update the *.nix files.

yeah, this part is not ideal. i should have pypi2nix[1] ready for
public use by the end of april.


> Builds verify hashes (well, md5's.. so, not worse than not validating), but aren't deterministic, so a little better than installing with pip from requirements.txt (and slightly worse than using peep)

there is guaranteed build determinism, meaning packages are build the
same way with the same dependencies. this holds bigger value then
actual binary determinism (byte-by-byte) which practicality is only
valuable for security reasons.


> Non-Python dependencies (ldap, mysql, etc.) are completely specified, so that's cool -- but I still feel like this is a solution in search of a problem.

i know this are hard to see benefits, but imagine that once you define
complete dependency try you can package that tree as
docker/AWS/virtualbox image or any other deployment format. and same
binaries are used. being able to do QA on binaries you're going to
deploy is a big advantage for me, also this is (or at least it should
be) many times a requirement for a continuous development system.


> We are on our own to handle any security upgrades in libraries we depend on, which are hidden from typical scanning tools. So shipping these Non-Python dependencies increases our burden vs. just apt-get install libmysql.

this part you misunderstood. instead of imperatively doing an update
(eg. running `apt-get update`) you update to newer version of nixpkgs
which contains latest bugfixes.

what nix makes very easy to do is that you can (but you're not
required to) update package from inside default.nix, not needing any
extra infrastructure or running any extra commands. use-case for this
would be that for some strange reason we need to depend on something
specific version of mysql.

another example would be that you can apply custom patches (until they
are included into upstream release) on the fly without doing any
"internal" release. a nice example for this would be
`PyHawk-with-a-single-extra-commit`[2]. seriously doing a release just
to include one simple commit? that is just silly.


[1] https://github.com/garbas/pypi2nix
[2] https://pypi.python.org/pypi/PyHawk-with-a-single-extra-commit

djmitche:
In light of the tech talk -- ship it!

PR #380, opened 2016-09-02
Bug 1299960 - Linkify URLs when one is not a sheriff

KWierso:
Totally untested since I don't want to set up an entire buildapi instance on my laptop at 1am, but I think this is what we would want if we wanted to do this. Maybe I need to reverse the order between linkifyBugs and linkifyURLs, dunno if they conflict with each other.

### DIFF
--- a/relengapi/blueprints/treestatus/static/tree.html
+++ b/relengapi/blueprints/treestatus/static/tree.html
@@ -12,8 +12,11 @@
                 <span class="{{tree.status|status2class}}">{{tree.status|uppercase}}</span>
             </h2><br/>
             <h2 class="treestatus" ng-if="tree.reason">Reason:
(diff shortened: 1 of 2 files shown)


catlee:
Yeah, I suspect linkifyUrls will match the URLs in the bug links, and re-linkify them.

Are there better ways of linkifying urls here rather than regexes?

KWierso:
Actually, looking back at this from a not-1am perspective, I don't think this will even do what we want. The linkifyURLs filter takes a url and turns it into the <a href="url">url</a>. It doesn't make a text node containing html content render as html.

garbas:
@KWierso i took over the ticket in bugzilla and will shedule to look at it next week. i will close this PR, bbut i will keep it in mind when working on a fix.

